PR #694, opened 2016-02-03
Custom directory layout

### LINKED ISSUE #565: Custom directory layout (initial support)
For instance, [Bedrock](https://roots.io/bedrock/) (the roots.io project) uses a folder structure like this:

```
├── composer.json
├── config
│   ├── application.php
│   └── environments
│       ├── development.php
│       ├── staging.php
│       └── production.php
├── vendor
└── web
    ├── app
    │   ├── mu-plugins
    │   ├── plugins
    │   ├── themes
    │   └── uploads
    ├── wp-config.php
    ├── index.php
    └── wp
```

> Note: Bedrock also uses custom configuration system so we cannot support it as a whole yet, however, the **structure** is something we possibly could.

We've also had similar request from e-labo.biz who use a similar structure but without the Composer files and with `index.php` and `wp-config.php` inside the `app` folder.
- _This issue is blocked by #630_

JanVoracek:
Resolves #565.

Structure of the project can be now a little bit non-standard.
- User can move WordPress into [separate directory](https://codex.wordpress.org/Giving_WordPress_Its_Own_Directory#Moving_a_Root_install_to_its_own_directory). User has to setup the `VP_PROJECT_ROOT` to the original root (`ABSPATH` before moving the WordPress) to do this.
- User can rename or move `wp-content`.
- User can rename or move the plugins dir (`wp-content/plugins`).
- User can rename or move the uploads dir (`wp-content/uploads`).

It is necessary to do these customizations before activating VersionPress! 

Related docs: [live](https://docs.versionpress.net/en/feature-focus/custom-project-structure), [PR](https://github.com/versionpress/docs/pull/41).

Reviewers:
- [x] @borekb

### DIFF
--- a/plugins/versionpress/tests/test-config.sample.neon
+++ b/plugins/versionpress/tests/test-config.sample.neon
@@ -1,7 +1,7 @@
 # Test configuration, see TestConfig
 
 selenium:
-    firefox-binary: C:/Path/To/FirefoxPortable/App/Firefox/firefox.exe
+    firefox-binary: /Users/johndoe/Path/To/Firefox.app/Contents/MacOS/firefox
     post-commit-wait-time: 500
 
 # Default values for all sites, may be overwritten by a specific site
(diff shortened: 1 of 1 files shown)


acki:
+1

PR #850, opened 2016-04-05
New command check-requirements

### LINKED ISSUE #844: Create a WP-CLI command for checking requirements
When a user upgrades / changes his server the site may break down because of wrong permissions / missing PHP extensions / missing Git etc. He uses the same database and moves the code by hand so there is no need for running the `restore-site` command. 

Maybe we should create a WP-CLI command just for checking the requirements. 

Here's the initial motivation: https://gitter.im/versionpress/versionpress?at=5701546ad9b73e635f68ca5a

What do you think @borekb, @octopuss?

octopuss:
added new `wp-cli` command `vp check-requirements` . PR resolves #844.

Please review:

- [x] @JanVoracek 

Thank you.

Depends on:

- [x] merged https://github.com/versionpress/docs/pull/30


octopuss:
@JanVoracek the test I thought fails only in this branch fails also in master (#886). So i think that this branch can be merged to `master`.

PR #913, opened 2016-04-12
Database and Storage does not match when VP is deactivated

### LINKED ISSUE #903: ActivationDeactivationTest::afterActivationTheFilesystemMatchDatabase fails
This test fails only when it's run in a tests of whole VersionPress are run and Selenium is used as "worker". The test does not fail when run separately, or using `wp-cli`.

octopuss:
Resolves #903 .

Reviewers:

- [x] @JanVoracek 

Thank you.


octopuss:
I will change code to use `$database->vpid` everywhere.

PR #919, opened 2016-04-13
Display warning and disable activation when VersionPress not build correctly

### LINKED ISSUE #800: rename ReplacerMethod `restoreOriginal` to "vp_" prefixed alternative
All methods and fields injected into `wpdb` by Versionpress should make clear that they are not part of original `wpdb` implementation. They should contain `vp_prefix`. Method `restoreOriginal` from `ReplacerMethods.src.php` does not match this condition.

Please fix it.

### LINKED ISSUE #880: Display warning when someone clones the repo and uses the plugin as-is
If someone just clones this repository a copies `plugins/versionpress` into his `wp-content` it fails on a warning. See versionpress/support#30 for an example.

VersionPress should print some nice error message 🙊  (maybe with a link to the GH releases).

octopuss:
Resolves #880 .

Please review also the discussion in issue #880 and add your ideas here.

Reviewers:

- [x] @borekb 
- [x] @JanVoracek 

Thank you.

### DIFF
--- a/plugins/versionpress/versionpress.php
+++ b/plugins/versionpress/versionpress.php
@@ -38,6 +38,11 @@ use VersionPress\VersionPress;
 
 defined('ABSPATH') or die("Direct access not allowed");
 
+if (!file_exists(__DIR__ . '/vendor/autoload.php')) {
+    register_activation_hook(__FILE__, 'disable_plugin_activation');
+    return;
+}
+
 require_once(__DIR__ . '/bootstrap.php');
 
 if (defined('WP_CLI') && WP_CLI) {
@@ -97,6 +102,10 @@ add_filter('automatic_updates_is_vcs_checkout', function () {
     $forceUpdate = UninstallationUtil::uninstallationShouldRemoveGitRepo(); // first commit was created by VersionPress
     return !$forceUpdate; // 'false' forces the update
 });
+
+function disable_plugin_activation() {
+    wp_die('<h1> VersionPress could not be activated</h1> <p>Seems that your copy of VersionPress was not build correctly. Please build it from <a href="https://github.com/versionpress/versionpress/blob/master/docs/Dev-Setup.md">sources</a> or <a href="https://github.com/versionpress/versionpress/releases/latest"> download from GitHub</a>');
+}
 
 function vp_register_hooks() {
     global $versionPressContainer;
--- a/plugins/versionpress/versionpress.php
+++ b/plugins/versionpress/versionpress.php
@@ -104,7 +104,7 @@ add_filter('automatic_updates_is_vcs_checkout', function () {
 });
 
 function disable_plugin_activation() {
-    wp_die('<h1> VersionPress could not be activated</h1> <p>Seems that your copy of VersionPress was not build correctly. Please build it from <a href="https://github.com/versionpress/versionpress/blob/master/docs/Dev-Setup.md">sources</a> or <a href="https://github.com/versionpress/versionpress/releases/latest"> download from GitHub</a>');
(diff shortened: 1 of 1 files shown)


borekb:
Copying (parts) of the discussion here where it belongs:

---

A combination of `register_activation_hook` and `wp_die` function is used, resulting in this:

screen shot 2016-04-13 at 9 18 22 am

User is directed to [*Development setup page*](https://github.com/versionpress/versionpress/blob/master/docs/Dev-Setup.md), or to GitHub's [latest release page](https://github.com/versionpress/versionpress/releases/latest) which redirects him to 2.2 version release. There is no easy way to link zip download directly. Although the [api](https://api.github.com/repos/versionpress/versionpress/releases/latest) contain link to zip file, I think that it is not worth to parse it somehow, and return the correct url to user. What do you think ?

(original: https://github.com/versionpress/versionpress/issues/880#issuecomment-209277357)

octopuss:
From what I've seen, users can obtain link to ZIP file using different approaches. Use `curl` and `grep` on backend or `jQuery` on frontend (call API and process JSON response).

borekb:
Overall, this is a great solution to the problem, let's just improve some details.

1. We should definitely provide a direct download link, there is no reason to complicate it for the users.
2. There is no need to point them to the source code or build steps, let's just give the a link to the correct ZIP file.
3. The text has some small grammar errors in it.

Overall, I'd update the text to this:

---

### VersionPress could not be activated

It seems that your copy of VersionPress was not built correctly. Please [download this ZIP file](#) and [install it again](#).

---

The first link would point to the right ZIP file and the other to the "Upload new plugin" page.

One thing I am not sure about it how to handle preview releases like the current beta. I feel the ZIP should be the latest stable one but in our current situation, it would be better to link to 3.0-beta. Maybe it could be something like "[download this ZIP file](#) (see [all releases](#))" or something like that.

borekb:
BTW, after this message, will the bad plugin still be installed? Can we uninstall it somehow automatically or should we instruct the user to do it manually?

borekb:
Another idea: auto-install the right version. WP-CLI [can install plugins from URLs](http://wp-cli.org/commands/plugin/install/) so we could use that code to auto-download the right VersionPress ZIP and install it for the user, maybe even silently. However, for now, the manual method is good enough.

JanVoracek:
Quick dev note:
To get the URL of the ZIP we need `php_openssl` extension and enabled `allow_url_fopen` in `php.ini`.

octopuss:
Or we can use other [GitHub tags API endpoint](https://api.github.com/repos/versionpress/versionpress/tags). I think with some simple javascript it can work.

borekb:
Ok, `php_openssl` etc. is too much hassle, let's stay with the manual download method (https://github.com/versionpress/versionpress/pull/919#issuecomment-209328372).

borekb:
Sorry I did not realize that even a simple JSON request will require `php_opnssl` etc. So let's just link to the `/releses` page where the user will be able to see both stable and preview releases and decide which one to download. Should be enough for now.

octopuss:
And what about proposed javascript solution ? I've created simple example that works. See https://jsfiddle.net/4zs76056/. It is only dependent on simple AJAX library providing XHR calls wrapping and can be added to VersionPress. Or we can use our own minimalistic implementation.

borekb:
That would work, yes. However, there are subtle nuances like do we want to link them to the latest beta or the sable release? Simple link to `/releases` will be preferable for now, and not many people should encounter this page anyway.

octopuss:
So the final message looks like the one below.

## VersionPress could not be activated
It seems that your copy of VersionPress was not built correctly. Please download [release ZIP file from GitHub](https://github.com/versionpress/versionpress/releases) and [install it again](#).

PR #922, opened 2016-04-13
PHP_BINARY fix

### LINKED ISSUE #876: Undo does not work for added postmeta
When I try to undo commit related to added postmeta, I get following error: `Error: Overwritten changes can not be reverted.`

Steps to reproduce:

1) Add any postmeta to any post: commit "Created postmeta AB for post CD".
2) Save (update) this post: commit "Updated post CD".
3) Try to undo first postmeta commit.

Related question: Should we commit every post update even if only `post_modified` was changed? It can lead to many commits without relevant values.

borekb:
Resolves #876.

PHP binary is now detected from `extension_dir` rather than `PHP_BINARY` which is not set for requests running through a web server. `extension_dir` is a workaround but should be quite reliable on Windows, and Windows is the only OS where we need this workaround (Linux and Mac OS use Bash merge driver).

Reviewers:

- [x] @JanVoracek 
- [x] @octopuss

### DIFF
--- a/plugins/versionpress/src/Git/MergeDriverInstaller.php
+++ b/plugins/versionpress/src/Git/MergeDriverInstaller.php
@@ -1,6 +1,7 @@
 <?php
 namespace VersionPress\Git;
 
+use Nette\Utils\Strings;
 use VersionPress\Utils\PathUtils;
 use VersionPress\Utils\StringUtils;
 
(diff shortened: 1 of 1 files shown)


plugins/versionpress/src/Git/MergeDriverInstaller.php
```diff
@@ -92,7 +93,19 @@ private static function installGitConfig($rootDir, $pluginDir, $driver) {
...
+
+            // Finding the PHP binary is a bit tricky because web server requests don't use the main PHP binary at all
+            // (they either use `mod_php` or call `php-cgi`). PHP_BINARY is only correct when the process
+            // is initiated from the command line, e.g., via WP-CLI when cloning.
+            //
+            // We'll only fix the path for Windows because on Linux and Mac OS, Bash driver is used most of time
+            // and the tests are started from command line so PHP_BINARY is fine there too.
+            $phpBinary = PHP_BINARY;
+            if (DIRECTORY_SEPARATOR == '\\' && !Strings::endsWith($phpBinary, 'php.exe')) {
+                $phpBinary = realpath(ini_get('extension_dir') . '/..') . '/php.exe';
```
octopuss: The code looks good, i am just not sure if function `ini_get` will work everywhere. From what I remember, it was sometimes disabled on hostings. But that's just my doubts.

octopuss: I see that we use it in our code without problems, so it is ok.

borekb: Good point, I didn't check that. If there's a more reliable way to get the extension path we can swap the implementation. @JanVoracek, any ideas?

JanVoracek: `ini_get` should be OK. `ini_set` is sometimes disabled.

PR #926, opened 2016-04-14
Old commits not ignored when searching

### LINKED ISSUE #874: Old commits not ignored when searching
Users can for example "Undo" commits from previous VP version, see attached screens.

Before search
before_search

After searching "post"
after_search

vasek17:
Resolves #874.

Reviewers:

- [ ] @pavelevap 
- [x] @JanVoracek

### DIFF
--- a/plugins/versionpress/src/Api/VersionPressApi.php
+++ b/plugins/versionpress/src/Api/VersionPressApi.php
@@ -201,16 +201,15 @@ class VersionPressApi {
 
         $initialCommitHash = $this->getInitialCommitHash();
 
-        $isChildOfInitialCommit = $this->gitRepository->wasCreatedAfter($commits[0]->getHash(), $initialCommitHash);
         $isFirstCommit = $page === 0;
 
         $result = array();
         foreach ($commits as $commit) {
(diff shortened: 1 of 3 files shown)


plugins/versionpress/src/Api/VersionPressApi.php
```diff
@@ -201,16 +201,15 @@ public function getCommits(WP_REST_Request $request) {
 
         $initialCommitHash = $this->getInitialCommitHash();
 
-        $isChildOfInitialCommit = $this->gitRepository->wasCreatedAfter($commits[0]->getHash(), $initialCommitHash);
         $isFirstCommit = $page === 0;
 
         $result = array();
         foreach ($commits as $commit) {
-            $isChildOfInitialCommit = $isChildOfInitialCommit && ($commit->getHash() !== $initialCommitHash);
+            $isChildOfInitialCommit = $this->gitRepository->wasCreatedAfter($commit->getHash(), $initialCommitHash);
```
JanVoracek: I think this has a quite huge performance impact. It runs `git log` for every single displayed commit. Is this the only suitable solution?

vasek17: Yes, I am aware of that. I'll try to optimize it a bit.

PR #931, opened 2016-04-15
Deactivation and activation of VersionPress keeps original VPIDs and files

### LINKED ISSUE #906: Update method that keeps the VPIDs
Currently, the only reliable method to update between VP releases is to deactivate the plugin, upload the new version and re-activate. This unfortunately makes a thick line between the current and previous installations of VersionPress – all connection to the past is lost, VPIDs are re-generated, INI files created from scratch, etc. This is because the INI format still changes between releases and migrations are hard, see #275.

Until we have proper migrations, or until the format stabilizes so that we don't really need to change it, there could be a simpler method for the meantime: keep the `wp_vp_id` table, and when re-generating INI files from scratch, use this table to assign the same VPIDs to the newly created entities. This would have two main benefits:

1. I'd be able to track history of e.g. a post between VP activations.
2. Changes in INI files are usually quite small so reusing the same file would save a lot of data in Git repository. (Currently, all the files are deleted and new ones created, which is a big waste if there is a vast amount of entities.)

JanVoracek:
Resolves #906.

When user deactivates and activates VersionPress, it uses original IDs where it's possible. It means that now it is possible to track history of e.g. post even after update of VersionPress. Technically, it was achieved by moving some of the logic from deactivation to uninstallation. For example, INI files are now only deleted on uninstallation.

The update process remains the same:

1. Deactivate VersionPress
2. Replace it with a new version (or simply call `wp plugin install <path to zip> --force`)
3. Activate VersionPress
4. Initialize VersionPress

Is this OK or should I create some WP-CLI command for update?

I also encountered a little problem: VersionPress cannot be updated using GUI only. WordPress refuses new ZIP if VersionPress is already installed. The user may click on the delete link and the history will be gone. 

Reviewers:
- [x] @borekb 
- [x] @octopuss


borekb:
WP-CLI command would be useful. It doesn't need to be too fancy, for example, something simple like

    wp vp update path-to-newer-versionpress.zip

would be enough. In the future, we should integrate with WP Upgrader and deprecate this temporary helper command.

Other than that, the code looks :+1: by me.

octopuss:
From my point of view, the code looks good too.

unclhos:
Oh PLEASE give me an update command :)

JanVoracek:
I pushed an implementation of the `wp vp update` command. Feel free to comment.

borekb:
This will be super-useful. Thanks.

plugins/versionpress/src/Cli/SilentUpgraderSkin.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace VersionPress\Cli;
+
+class SilentUpgraderSkin extends \WP_Upgrader_Skin {
```
octopuss: Please add simple note what made you use this silent skin here or in `vp:update()` method.

JanVoracek: Done.

PR #932, opened 2016-04-15
Enable column ignoring

### LINKED ISSUE #902: Enable ignoring of specific database column
VersionPress is currently able to ignore some entities according to their content, i.e., it can ignore database rows. It is not capable to ignore only part of an entity, i.e., a database column of it. It is done for each mapped entity in its Storage.

Information about ignored columns (those we do not want to version see #901 or those with computed value) should be written in `wordpress-schema.yml` in more general way.

octopuss:
Resolves #902 .

How to define ignored columns is described in file `schema-readme.md`

Reviewers:

- [x] @JanVoracek 
- [x] @borekb

### DIFF
--- a/plugins/versionpress/src/Database/EntityInfo.php
+++ b/plugins/versionpress/src/Database/EntityInfo.php
@@ -136,9 +136,7 @@ class EntityInfo {
     private $frequentlyWritten = array();
 
     private $ignoredEntities = array();
-
-    public $hasIgnoredColumns = false;
-
+    
     private $ignoredColumns = array();
 
     /**
@@ -226,7 +224,6 @@ class EntityInfo {
                     $this->ignoredColumns[array_keys($column)[0]] = substr(array_values($column)[0], 1);
                 }
             }
-            $this->hasIgnoredColumns = true;
         }
 
     }
--- a/plugins/versionpress/src/Database/EntityInfo.php
+++ b/plugins/versionpress/src/Database/EntityInfo.php
@@ -218,31 +218,21 @@ class EntityInfo {
         }
 
         if (isset($schemaInfo['ignored-columns'])) {
-            $this->ignoredColumns = $schemaInfo['ignored-columns'];
+
+            foreach ($schemaInfo['ignored-columns'] as $column) {
+                if (is_string($column)) {
+                    $this->ignoredColumns[$column] = function () {}; // if column does not have any compute function we create 'NOOP' function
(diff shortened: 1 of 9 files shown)


plugins/versionpress/src/Database/schema-readme.md
```diff
@@ -114,3 +114,11 @@ It is possible to ignore some entities (don't save them into INI files). You can
...
         - 'option_name: _*'
         - 'option_name: siteurl'
+
+## Ignoring entities columns
+
+It is possible to ignore some entity columns (don't save them into INI files) and have their values computed if necessary. You can just write column name or combination of column name and function which should be called. Function is used for computing column value during synchronization process. Instance of `VersionPress\Database\Database` is passed as an argument to a function call.
+
+    ignored-columns:
+        - comment_count: '@\VersionPress\Synchronizers\PostsSynchronizer::fixCommentCounts'
+        - 'user_activation_key'
```
JanVoracek: Single quotes are not necessary (also in the schema file).

plugins/versionpress/src/Database/wordpress-schema.yml
```diff
@@ -13,6 +13,8 @@ post:
     #  - Attachments without post_title
     #  - Updating draft via AJAX
     #  - Any changes saved by clicking "Preview"
+  ignored-columns:
+    - comment_count: '@\VersionPress\Synchronizers\PostsSynchronizer::fixCommentCounts'
```
JanVoracek: I'm not sure about this notation. It looks simple but maybe it can be confusing. In one case is the column name a value, in other case it's a key and value is a function. What do think, @borekb?

borekb: It is both simple and slightly confusing :) Can't think of a better alternative though, let's stick with the current format.

borekb:
The format and schema-readme seem fine to me, ticking myself off.

plugins/versionpress/src/Database/EntityInfo.php
```diff
@@ -136,6 +137,10 @@ class EntityInfo {
 
     private $ignoredEntities = array();
 
+    public $hasIgnoredColumns = false;
```
JanVoracek: Is this field used anywhere?

octopuss: @JanVoracek I've removed it from the code.

PR #942, opened 2016-04-18
DbAsserter is more strict when comparing stored values against database

### LINKED ISSUE #909: DbAsserter should fail if database contains data different from files
In #900 we have revealed problem, that some columns are not inserted by VersionPress and their values are filled by columns defaults set in DDL. This behavior causes inconsistency between database and ini files. Current implementation of DbAsserter does not mark it as a failure (inconsistency is skipped https://github.com/versionpress/versionpress/blob/db645ea95a9f4b4016878b453ccedb0779ca3e0d/plugins/versionpress/tests/Utils/DBAsserter.php#L123-L123)

We should change this behavior to be more strict, and warn/fail when files does not contain some of database columns. 

DbAsserter itself should be also reviewed, if there are other places for improvement.

octopuss:
Resolves #909.

`DbAsserter.php` now compares each column from database with value returned from `Storage`. Before this change if column from database was not found in storage it was skipped from any other processing. Implementing this issue has also 2 side effects : 
- ignoring of option `theme_switched`
- `false` values sent to *INSERT* or *UPDATE* query are now transformed to `""` (empty string)

Reviewers:

- [x] @JanVoracek 
- [x] @borekb

### DIFF
--- a/plugins/versionpress/src/Database/WpdbMirrorBridge.php
+++ b/plugins/versionpress/src/Database/WpdbMirrorBridge.php
@@ -50,9 +50,9 @@ class WpdbMirrorBridge {
         $entityName = $entityInfo->entityName;
         $data = $this->vpidRepository->replaceForeignKeysWithReferences($entityName, $data);
 
-        array_walk($data, function (&$key, $value) {
+        array_walk($data, function (&$value, $key) {
             if ($value === false) {
-                $key = '';
+                $value = '';
             }
         });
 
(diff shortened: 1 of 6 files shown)


borekb:
I like this change a lot, it makes the schema more explicit and the tests more complete / reliable.

plugins/versionpress/src/Database/WpdbMirrorBridge.php
```diff
@@ -49,6 +49,13 @@ function insert($table, $data) {
 
         $entityName = $entityInfo->entityName;
         $data = $this->vpidRepository->replaceForeignKeysWithReferences($entityName, $data);
+
+        array_walk($data, function (&$key, $value) {
+            if ($value === false) {
+                $key = '';
```
JanVoracek: This is weird.

1. The signature of callback of `array_walk` is ($value, $key).
2. Variable `$value` (which represents key in this case) cannot be `false`.

octopuss: Done.

PR #946, opened 2016-04-19
Separate gulp task for Selenium

### LINKED ISSUE #936: Enable running Selenium as a separate gulp task
Gulp task `gulp run-tests` now downloads and starts Selenium and runs the tests. It would be handy to run the Selenium alone (for running individual tests).

JanVoracek:
Resolves #936.

I split the original `run-tests` into two tasks – one for starting Selenium server, the second one for running the tests.

Reviewers:
- [x] @borekb 
- [x] @octopuss


borekb:
This is a great improvement. Just please add a small note about it to `Testing.md`.

Issue #847, opened 2016-04-04
Rename default environment from `master` to `default`


borekb:
Since https://github.com/versionpress/versionpress/issues/459, we track environments in which commits originated. On any clone, that clone name is used as an environment name, e.g., `staging` or `bobs-clone`. However, the original site (without the user manually configuring the `VP_ENVIRONMENT` constant) is always called `master`.

Master is commonly a branch name, not an environment name. We should reconsider how to call that env, or even possibly force the user to always pick a default env name if we can't come up with something generic enough.

borekb:
Moving this to 3.0, we need to get rid of this sooner rather than later.

As always, naming is hard :) Some ideas of what the original environment could be called:

- **default** – the obvious name but is it OK to use it as a standalone noun?
- **origin** – good name but could be confusing in Git context, for example, "the origin environment has a Git remote origin ..."
- **base** – short and quite good
- **main**, **trunk** and similar – they suggest that the environment is significant which might be misleading. For instance, imagine two environments, `main` and `production` – which is the "really main" one?

borekb:
Let's go with *default* for 3.0. We might provide a text field in the Initializer in the future so that the user could pick a better name for his/her current env.

The change from `master` to `default` will only need to happen on the UI level; files, commits etc. are not affected by this.

borekb:
Another thing that should be updated is the `vp clone` command. It [currently][1] creates the "origin" pointer back to the original site which should be updated to `default`. We should avoid using "origin" altogether because that would commonly point to a remote repo, not a local clone.


  [1]: https://github.com/versionpress/versionpress/blob/893dfb4b3f627110be50dfa8d071e06306e298a0/plugins/versionpress/src/Cli/vp.php#L437-L437

borekb:
Closely related to this is PR #955 (issue #951). After that PR is merged, what's left of this ticket is to remove references to `master` from places like [ChangeInfoEnvelope][1].


  [1]: https://github.com/versionpress/versionpress/blob/966d7737b8ae2cf68e2c6f23f5724aab3c555252/plugins/versionpress/src/ChangeInfos/ChangeInfoEnvelope.php#L23-L23

borekb:
A couple of `master` references in comments that the PR #968 missed:

- [VersionPress.php#L22](https://github.com/versionpress/versionpress/blob/013b0c5ade203ac038617b023127519d5449c802/plugins/versionpress/src/VersionPress.php#L22-L22)
[...]

PR #955, opened 2016-04-20
Set environment name before activation

### LINKED ISSUE #951: Offer option to name the default environment via UI before initialization
Clones created from the original sites always have names ('wp vp clone **--name=xyz**') but the original one doesn't – we call it **default** (previously master, see #847). When browsing the environments – see https://github.com/versionpress/versionpress/pull/858 – the name `default` doesn't look too good next to better env. names like `staging`, `dev` etc. We should give the user an opportunity to provide a better name for the default environment, for example, some users might call their original environment `production` or something like that.

borekb:
Resolves #951.

There's a new piece of UI on the pre-activation screen:

name-environment-pre-activation

The entered environment name is validated slightly in JavaScript (quick alert, nothing too fancy) and then on the server-side. If server-side validation fails, `default` is used as an env name (see #847).

Reviewers:

- [x] @JanVoracek 
- [x] @octopuss

### DIFF
--- a/plugins/versionpress/admin/inc/activationPanel.php
+++ b/plugins/versionpress/admin/inc/activationPanel.php
@@ -7,13 +7,6 @@ use VersionPress\Utils\RequirementsChecker;
 
 ?>
 
-<style>
-    h4 {
-        margin-top: 30px;
-        margin-bottom: 10px;
-    }
-</style>
-
 <script>
 
 jQuery(document).ready(function($) {
@@ -112,7 +105,7 @@ jQuery(document).ready(function($) {
                         E.g., <code>production</code>, <code>dev</code> or <code>my-machine</code>. This will help you identify this environment later.
 
                         <div style="margin-top: 10px;">
-                            <label for="envname"><?php _e('Environment name:') ?></label>
+                            <label for="envname">Environment name:</label>
                             <input name="envname" type="text" id="envname" value="default" />
 
                         </div>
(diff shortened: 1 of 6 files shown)


plugins/versionpress/admin/inc/activationPanel.php
```diff
@@ -79,6 +104,25 @@
...
 
+                <h4>Name your environment</h4>
+
+                <ul>
+                    <li>
+                        <span class="dashicons dashicons-info icon" style="color: #5b9dd9; font-size: 1.2em; left: -28px;"></span>
+                        E.g., <code>production</code>, <code>dev</code> or <code>my-machine</code>. This will help you identify this environment later.
+
+                        <div style="margin-top: 10px;">
+                            <label for="envname"><?php _e('Environment name:') ?></label>
```
octopuss: Localization function call is not necessary right now.

plugins/versionpress/admin/inc/activationPanel.php
```diff
@@ -7,6 +7,31 @@
 
 ?>
 
+<style>
+    h4 {
+        margin-top: 30px;
+        margin-bottom: 10px;
+    }
+</style>
```
octopuss: Move styles into `styles.css`.

plugins/versionpress/admin/inc/activationPanel.php
borekb: Done.

plugins/versionpress/admin/inc/activationPanel.php
borekb: Done.

PR #963, opened 2016-04-21
Code style changed to PSR-2 and added Code Sniffer

### LINKED ISSUE #629: Check codestyle
Sometimes the code isn't in accordance with the [PSR-2](http://www.php-fig.org/psr/psr-2/) standard – mostly we're violating the "brace on the next line" rule on classes and their methods. We should fix it.

Also would be great to have the style checker on our CI.

JanVoracek:
Resolves #629.

VersionPress now meets the PSR-2 standard and uses new PHP features. There's a lot of modified files. 
Typical changes are:
 - brackets on new line,
 - long lines split to multiple lines,
 - arrays converted to short syntax,
 - using class name constant (`Class::class`).

Also, the project files are changed so I suggest to run `init-dev` and `idea` tasks.

Reviewers:
- [x] @octopuss


plugins/versionpress/admin/deactivate.php
```diff
@@ -30,26 +32,30 @@ function _vp_button($label, $action, $type = "delete", $cssClass = "") {
...
-
     <div class="error below-h2">
         <p>You are about to deactivate VersionPress. Be aware that:</p>
 
         <ul>
 
             <li>
                 <span class="icon icon-notification warning-color"></span>
-                <span class="vp-highlight-text">Later installations of VersionPress <strong>will not be able to undo or rollback changes</strong> done by this installation.</span> The changes are still technically in the repository if you want to inspect them using e.g. the command line tools etc.
+                <span class="vp-highlight-text">Later installations of VersionPress <strong>will not be able to undo or
```
borekb: Hard wraps in HTML usually make the editing less pleasant, maybe we could exclude these "PHP files which are essentially HTML files with a little bit of code" from code sniffing. We have only a few of them.

plugins/versionpress/.htaccess
```diff
@@ -1,10 +1,10 @@
 # Apache 2.2
 <IfModule !mod_authz_core.c>
-	Order Deny,Allow
-	Deny from all
+    Order Deny,Allow
```
borekb: Even .htaccess is now PSR-2 :) :+1:

plugins/versionpress/admin/deactivate.php
```diff
@@ -16,10 +16,12 @@
  * @param string $type See `submit_button()`'s $type parameter
  * @param string $cssClass Optional CSS class to be used for the button
  */
-function _vp_button($label, $action, $type = "delete", $cssClass = "") {
-    echo "<form action='" . esc_url(admin_url('admin-post.php')) . "' method='post' class='" . esc_attr($cssClass) . "'>";
+function _vp_button($label, $action, $type = "delete", $cssClass = "")
+{
+    echo "<form action='" . esc_url(admin_url('admin-post.php')) .
+        "' method='post' class='" . esc_attr($cssClass) . "'>";
```
borekb: These are the situations where I feel like the standard is more terrorizing us than helping. 123 chars line length is fine.

I've checked [PSR-2](http://www.php-fig.org/psr/psr-2/) to confirm that 120 is indeed the hard limit but it's not:

> There MUST NOT be a hard limit on line length; the soft limit MUST be 120 characters; lines SHOULD be 80 characters or less.

From [1. Overview](http://www.php-fig.org/psr/psr-2/#1-overview) and [2.3 Lines](http://www.php-fig.org/psr/psr-2/#2-3-lines). This will make long HTML lines and similar stuff much easier.

borekb:
The diff is really huge so I'll copy some of the main points of discussion here:

- I think we should not treat 120 chars as the hard limit for line length. PSR-2 states "there MUST NOT be a hard limit on line length" and for example paragraphs of HTML are better written as single lines with soft wraps in the editor. (The discussion started in [this inline comment](https://github.com/versionpress/versionpress/pull/963/files#diff-69ddd57af8042d4a3f0d6e1e20ba1451).)

plugins/versionpress/src/Initialization/ReplacerMethods.src.php
```diff
@@ -91,15 +96,16 @@ public function delete($table, $where, $where_format = null) {
...
+    {
         global $versionPressContainer;
         if ($versionPressContainer == null || $this->vp_use_original_query_method) {
             return $this->__wp_query($query);
         }
 
         $r = null;
 
-        /** @var SqlQueryParser $sqlQueryParser */
+        /** @var \VersionPress\Database\SqlQueryParser $sqlQueryParser */
```
borekb: Is this actually an important fix?

borekb:
Ok I've done the review, apart form long lines in HTML and some other rare cases, there are many nice improvements (e.g., [SystemInfo.php](https://github.com/versionpress/versionpress/blob/bd898332523ec68097315f3de7cf2e385537d208/plugins/versionpress/src/Utils/SystemInfo.php#L169-L169)).

Special props for making `versionpress.php` pass the checks :)

plugins/versionpress/src/Initialization/ReplacerMethods.src.php
JanVoracek: It's not important as it's just a comment. But it wasn't valid class in the context of replaced `wp-db.php`.

PR #964, opened 2016-04-21
Schema README review

### LINKED ISSUE #933: Review schema-readme.md
There are some typos and smaller things to review in `schema-readme.md`, when all PRs are merged let's do a quick review before 3.0 is released.

borekb:
Resolves #933.

Only stylistic updates, no technical change, but quite a few changes.

Reviewers:

- [x] @JanVoracek 
- [x] @octopuss

### DIFF
--- a/plugins/versionpress/src/Database/schema-readme.md
+++ b/plugins/versionpress/src/Database/schema-readme.md
@@ -182,7 +182,7 @@ mn-references:
   ~<junction_table_name>.<column_name>: <foreign_entity_name>
 ```
 
-Virtual references are not stored in INI files but the relationships are checked during reverts. For example, when restoring a post that references a category that no longer exists, this operation will fail.
+Virtual references are not stored in INI files but the relationships are checked during reverts. For example, when a revert would delete a category (revert of `term_taxonomy/create`) and there is some post referencing it, the operation would fail.
 
 
 ### Parent references
@@ -212,7 +212,7 @@ entity:
       interval: 5min
 ```
 
-The values in the `frequently-written` array can either be strings which are then interpreted as  queries, or objects with `query` and `interval` keys. The interval is parsed by the `strtotime()` PHP function and the default value is one hour.
+The values in the `frequently-written` array can either be strings which are then interpreted as queries, or objects with `query` and `interval` keys. The interval is parsed by the `strtotime()` PHP function and the default value is one hour.
 
 
 ## Ignoring entities
--- a/plugins/versionpress/src/Database/schema-readme.md
+++ b/plugins/versionpress/src/Database/schema-readme.md
@@ -212,7 +212,7 @@ entity:
       interval: 5min
 ```
 
-The values in the `frequently-written` array can either be strings which are then interpreted as queries, or objects with `query` and `interval` keys. The interval is parsed by the `strtotime()` PHP function and the default value is one hour.
+The values in the `frequently-written` array can either be strings which are then interpreted as queries, or objects with `query` and `interval` keys. Queries use the same syntax as search / filtering in the UI, with some small differences like that the date range operator cannot be used but overall, the syntax is pretty intuitive. The interval is parsed by the `strtotime()` PHP function and the default value is one hour.
 
 
 ## Ignoring entities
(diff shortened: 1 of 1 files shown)


plugins/versionpress/src/Database/schema-readme.md
```diff
@@ -1,124 +1,244 @@
...
+One entity is considered "master" which is kind of arbitrary (technically, they are equal) but here, we decided that posts will store tags and categories, not the other way around. INI files of posts will store the references.
+
+References can also be prefixed with a tilde (`~`) which makes them virtual:
+
+```
+mn-references:
+  ~<junction_table_name>.<column_name>: <foreign_entity_name>
+```
+
+Virtual references are not stored in INI files but the relationships are checked during reverts. For example, when restoring a post that references a category that no longer exists, this operation will fail.
```
JanVoracek: This is an example of standard M:N reference. A virtual reference works, for example, in a situation, when a revert would delete a category (revert of `term_taxonomy/create`) and there is some post referencing it.

plugins/versionpress/src/Database/schema-readme.md
```diff
@@ -1,124 +1,244 @@
...
+```
+entity:
+  frequently-written:
+    - 'column_name: value'
+    - query: 'column1_name: value1 column2_name: value2'
+      interval: 5min
+```
 
-The interval is parsed by PHP function `strtotime`, so it can be whatever the function takes.
+The values in the `frequently-written` array can either be strings which are then interpreted as  queries, or objects with `query` and `interval` keys. The interval is parsed by the `strtotime()` PHP function and the default value is one hour.
```
JanVoracek: Double space in `interpreted as  queries`

plugins/versionpress/src/Database/schema-readme.md
borekb: Thanks, fixed.

plugins/versionpress/src/Database/schema-readme.md
borekb: Well spotted :) Fixed.

plugins/versionpress/src/Database/schema-readme.md
```diff
@@ -1,124 +1,244 @@
...
+
+References can also be prefixed with a tilde (`~`) which makes them virtual:
+
+```
+mn-references:
+  ~<junction_table_name>.<column_name>: <foreign_entity_name>
+```
+
+Virtual references are not stored in INI files but the relationships are checked during reverts. For example, when a revert would delete a category (revert of `term_taxonomy/create`) and there is some post referencing it, the operation would fail.
+
```
octopuss: Once again emphasize that `junction_table_name` is without table prefix.

plugins/versionpress/src/Database/schema-readme.md
```diff
@@ -1,124 +1,244 @@
...
+```
+entity:
+  frequently-written:
+    - 'column_name: value'
+    - query: 'column1_name: value1 column2_name: value2'
+      interval: 5min
+```
 
-The interval is parsed by PHP function `strtotime`, so it can be whatever the function takes.
+The values in the `frequently-written` array can either be strings which are then interpreted as queries, or objects with `query` and `interval` keys. The interval is parsed by the `strtotime()` PHP function and the default value is one hour.
```
octopuss: I think that the `query` clause should be explained a little bit more. Maybe @JanVoracek will add examples and explanation, how the query is applied to the entity. e.g. that column names in `column1_name: value1 column2_name: value2` are matched using AND/OR, and what are other possibilities (regex, etc.)

plugins/versionpress/src/Database/schema-readme.md
borekb: Done.

plugins/versionpress/src/Database/schema-readme.md
borekb: I was thinking the same. Do queries use the same engine as our searching / filtering functionality in VP 3.0? If so, I'd just add a note about that.

JanVoracek: More or less. They don't support range search for date and each key (DB column) can appear at most once.

borekb: Ok, I've added a brief description of the query language.

PR #985, opened 2016-04-26
Activation on PHP < 5.6 now displays nice error message

### LINKED ISSUE #370: Better error message for PHP lower than 5.6
*This is an imported issue, original reference WP-370, created on 27 Apr 2015.*

It can be done like [this](http://wordpress.stackexchange.com/q/76007/12248). There should be only PHP 5.2 compatible code in `versionpress.php`.

JanVoracek:
Resolves #370.

Everything from `versionpress.php` is now in `setup-hooks.php`. Main plugin file now contains only PHP 5.2 compatible checks and require of the original code.

Reviewers:
- [x] @borekb 
- [x] @octopuss 
- [x] @JanVoracek


plugins/versionpress/versionpress.php
```diff
@@ -10,972 +10,38 @@
...
-
-        $translations = wp_get_available_translations();
-        return isset($translations[$code])
-            ? $translations[$code]['native_name']
-            : 'English (United States)';
+        wp_die('<h1>VersionPress could not be activated</h1>
+            <p>
+                You are using an unsupported version of PHP. We recommend using one of the
+                <a href="http://php.net/supported-versions.php">actively supported</a>.
+            </p>');
```
octopuss: I think that the message should inform user that the PHP version on his server is unsupported by VersionPress itself. Version he is using may be fine for his other business.

borekb:
Change implemented by https://github.com/versionpress/versionpress/pull/985/commits/d23e7cb77c91a5ad49bf4b9c3eddf9d2315e2a8e looks like this:

2016-04-26 13_16_53-clipboard

There is:

- an inline notice in a table row
- red admin notice on `plugins.php`
- ... and only on `plugins.php` – VersionPress simply does nothing on all other screens.

Intentionally, the only way to get rid of the admin notice is to either upgrade PHP or disable VersionPress. If PHP is upgraded while VersionPress is kept activated, it will work fine – the red messages will just go away and the standard "VersionPress is activated but not yet tracking the site" message will pop up.

PR #998, opened 2016-04-26
Fixed test of WpConfigSplitter

### LINKED ISSUE #953: Put a comment to wp-config.common.php
`wp-config.common.php` is empty most of the time and users might be wondering what this file is. We should put a small comment there.

octopuss:
Resolves #953 .

Reviewers:

- [x] @JanVoracek 

thank you.

### DIFF
--- a/plugins/versionpress/tests/Unit/WpConfigSplitterTest.php
+++ b/plugins/versionpress/tests/Unit/WpConfigSplitterTest.php
@@ -101,9 +101,6 @@ define( 'AUTOMATIC_UPDATER_DISABLED', true );
 DOC;
 
     private $commonConfig = <<<'DOC'
-<?php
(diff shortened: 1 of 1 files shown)


plugins/versionpress/tests/Unit/WpConfigSplitterTest.php
```diff
@@ -102,6 +102,8 @@ class WpConfigSplitterTest extends \PHPUnit_Framework_TestCase
 
     private $commonConfig = <<<'DOC'
 <?php
+// Configuration common to all VersionPress environments, included from `wp-config.php`.
+// Learn more at https://docs.versionpress.net/en/getting-started/configuration
```
borekb: It would be safer to load this from the template file, not hardcode it here.

PR #1008, opened 2016-04-27
All paths in `ruleset.xml` are now relative

### LINKED ISSUE #1007: Make ruleset.xml portable
Exclude paths in `ruleset.xml ` should be relative which is best for portability (for example, prevents unpleasant surprises when PHP Code Sniffer is run in an unexpected path).

borekb:
Resolves #1007.

By default, PHP Code Sniffer runs exclude-patterns against full paths which might be problematic e.g. when PhpStorm copies the project to a temporary location to run `phpcs` on it – suddenly, "temp" is in path and all the checks are skipped. With this change, all the paths are relative to where `phpcs` is executed, which is in the VersionPress plugin root (where `ruleset.xml` lives) both locally and on Travis.

Reviewers:

- [x] @JanVoracek

### DIFF
--- a/plugins/versionpress/ruleset.xml
+++ b/plugins/versionpress/ruleset.xml
@@ -3,20 +3,20 @@
     <rule ref="PSR2"/>
 
     <rule ref="PSR1.Methods.CamelCapsMethodName.NotCamelCaps">
-        <exclude-pattern>*Worker.php</exclude-pattern>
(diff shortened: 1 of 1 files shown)


octopuss:
It seems to work on my MacBook.

JanVoracek:
Finally it looks like XML :))

PR #1011, opened 2016-04-28
VP prefix is stripped when commit message is displayed to user

### LINKED ISSUE #1009: Hide [VP] commit message prefix in VersionPress GUI
In several cases (see #971 or image below) user can see technical prefix `[VP]` which is added to each commit message created by VersionPress. As seen in picture below, it can be quite confusing because revert commit message is different than original commit message. `[VP]` prefix should be hidden.

screen shot 2016-04-27 at 4 02 02 pm

octopuss:
Resolves #1009.

Reviewers:

- [x] @JanVoracek

### DIFF
--- a/plugins/versionpress/bootstrap.php
+++ b/plugins/versionpress/bootstrap.php
@@ -54,11 +54,11 @@ if (!defined('VERSIONPRESS_REQUIRE_API_AUTH')) {
     define('VERSIONPRESS_REQUIRE_API_AUTH', true);
 }
 
-if (!defined('VP_COMMIT_MESSAGE_PREFIX')) {
+if (!defined('VERSIONPRESS_COMMIT_MESSAGE_PREFIX')) {
     /**
      * Prefix added in front of each commit message subject created by VersionPress
      */
-    define('VP_COMMIT_MESSAGE_PREFIX', '[VP] ');
+    define('VERSIONPRESS_COMMIT_MESSAGE_PREFIX', '[VP] ');
 }
 
 require_once(VERSIONPRESS_PLUGIN_DIR . '/vendor/autoload.php');
--- a/plugins/versionpress/bootstrap.php
+++ b/plugins/versionpress/bootstrap.php
@@ -54,6 +54,13 @@ if (!defined('VERSIONPRESS_REQUIRE_API_AUTH')) {
     define('VERSIONPRESS_REQUIRE_API_AUTH', true);
 }
 
+if (!defined('VP_COMMIT_MESSAGE_PREFIX')) {
+    /**
+     * Prefix added in front of each commit message subject created by VersionPress
+     */
+    define('VP_COMMIT_MESSAGE_PREFIX', '[VP] ');
+}
+
 require_once(VERSIONPRESS_PLUGIN_DIR . '/vendor/autoload.php');
 require_once(VERSIONPRESS_PLUGIN_DIR . '/versionpress-functions.php');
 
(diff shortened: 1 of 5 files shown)


plugins/versionpress/bootstrap.php
```diff
@@ -54,6 +54,13 @@
     define('VERSIONPRESS_REQUIRE_API_AUTH', true);
 }
 
+if (!defined('VP_COMMIT_MESSAGE_PREFIX')) {
```
JanVoracek: We use prefix `VP_` for public constants, `VERSIONPRESS_` for internal constants. I'm not sure if this should be a public constant.

octopuss: Is it just semantic distinguishing or is there also mechanism which handles accessibility of those constants ? I have no problem with making it internal. Just asking :)

plugins/versionpress/src/Git/CommitMessage.php
```diff
@@ -57,6 +57,19 @@ public function getSubject()
     }
 
     /**
+     * Returns commit message subject without prefix set in constant VP_COMMIT_MESSAGE_PREFIX
+     * 
+     * @return string
+     */
+    public function getUnprefixedSubject()
```
JanVoracek: I think stripping the prefix from the commit message is not a responsibility of the `CommitMessage` class. If `GitRepository` adds the prefix, it should also remove it (probably in the `log` method).

octopuss: From my point of view it is same service of `CommitMessage` class as its method `getVersionPressTags`. Tags are also not added as constructor args to `CommitMessage` instance but the instance itself returns them.

borekb: If you want 3rd opinion, I'd agree more with @JanVoracek. Only if it's too hard to do, technically, I'd go with this (`CommitMessage`) solution.

JanVoracek: @octopuss: Yes, tags should be refactored.

octopuss: Guys I don't agree that `GitRepository` is the right place for handling how the message should be presented to user. It should just obtain commit messages from underlaying GIT repository as is without manipulating it. 

If I strip it right in the `GitRepository` now, we could get rid of the prefix at all, because there will be no benefits of having it in commit messages. In proposed implementation, we will need to do some work to obtain the prefix, because it will be stripped by default. I understand that stripping can be optional but it will add some noise there.

For me letting `CommitMessage` class provide option to return it's properties in filtered or unfiltered way is better than noising `GitRepository`. For me is this approach a little bit similar to Java's `.toString()` methods.

octopuss: @JanVoracek: if result of will be removal of those getters, it will make our domain more anemic, but ok.

borekb: It's not a good time to argue about this (just before the 3.0 release), and if there are refactorings around these things anyway, let's keep the code the way it is (in `CommitMessage`) and reconsider in the future. We need to review the whole VP prefix thing anyway.

JanVoracek: > because there will be no benefits of having it in commit messages

We don't use it anyway. It's just for users who use other tool than VersionPress to see the commits.

OK, we will change this later.

PR #1016, opened 2016-04-28
Fixed using __FILE__ constant in setup-hooks.php

### LINKED ISSUE #370: Better error message for PHP lower than 5.6

JanVoracek:
Resolves #370.

Reviewers:
- [x] @octopuss


octopuss:
Whoa, that was quick 👍

PR #1020, opened 2016-04-28
Happy blogging

### LINKED ISSUE #1010: Better place detection in `wp vp config` (a.k.a. "happy blogging" issue)
We should not depend on the `/* That's all, stop editing! Happy blogging. */` comment as it is subject to localization, see e.g. #999.

JanVoracek:
Resolves #1010.

The algorithm now tries to find a one-line comment containing exclamation mark (*stop editing!*, *ne touchez pas à ce qui suit !*, *Schluss mit dem Bearbeiten!*, *Это всё, дальше не редактируем. Успехов!*). If there is no such comment it tries to find `if ( !defined('ABSPATH') )` or `require_once(ABSPATH . 'wp-settings.php');`.

Reviewers:
- [x] @octopuss


octopuss:
Constant is present after deactivation, but it is no big deal. I will create issue for that in milestone 3.0.1

PR #1042, opened 2016-05-11
Remove VP constants on deactivation

### LINKED ISSUE #1021: VP constants are present in wp-config.php after plugin deactivation
Steps to reproduce: 
- Install WordPress
- in `wp-config.php` replace line`/* That's all, stop editing! Happy blogging. */` with custom constant e.g. `define('my_constant', true);`
- activate and deactivate VersionPress

Lines below remains in file.
```

/** Absolute path to the WordPress directory. */
define('VP_ENVIRONMENT', 'default');

```

Related issue #1020

octopuss:
Resolves #1021 .

VersionPress constants are removed on uninstall. Also performed refactor of public constants (`VP_` prefixed)

Reviewers:

- [x] @JanVoracek 

Thank you

### DIFF
--- a/plugins/versionpress/setup-hooks.php
+++ b/plugins/versionpress/setup-hooks.php
@@ -659,11 +659,6 @@ function vp_admin_post_confirm_deactivation()
         FileSystem::remove(VERSIONPRESS_ACTIVATION_FILE);
     }
 
-    $wpConfigPath = \VersionPress\Utils\WordPressMissingFunctions::getWpConfigPath();
-    \VersionPress\Utils\WpConfigEditor::removeVersionPressConstants([
-        $wpConfigPath,
-        dirname($wpConfigPath . '/' . WpConfigSplitter::COMMON_CONFIG_NAME),
-    ]);
     global $versionPressContainer;
     /** @var Committer $committer */
     $committer = $versionPressContainer->resolve(VersionPressServices::COMMITTER);
--- a/plugins/versionpress/setup-hooks.php
+++ b/plugins/versionpress/setup-hooks.php
@@ -659,6 +659,11 @@ function vp_admin_post_confirm_deactivation()
         FileSystem::remove(VERSIONPRESS_ACTIVATION_FILE);
     }
 
+    $wpConfigPath = \VersionPress\Utils\WordPressMissingFunctions::getWpConfigPath();
+    \VersionPress\Utils\WpConfigEditor::removeVersionPressConstants([
+        $wpConfigPath,
+        WpConfigSplitter::COMMON_CONFIG_NAME,
+    ]);
     global $versionPressContainer;
     /** @var Committer $committer */
     $committer = $versionPressContainer->resolve(VersionPressServices::COMMITTER);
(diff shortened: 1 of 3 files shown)


plugins/versionpress/src/Initialization/WpConfigSplitter.php
```diff
@@ -35,9 +35,10 @@ public static function split($wpConfigPath, $commonConfigName = self::COMMON_CON
         $commonConfigLines = is_file($commonConfigPath)
             ? file($commonConfigPath)
             : file(__DIR__ . '/wp-config.common.tpl.php');
-
+        $constantsClass = new ReflectionClass(PublicConstants::class);
+        $vpPublicConstants = array_values($constantsClass->getConstants());
         // https://regex101.com/r/zD3mJ4/2
-        $constantsForRegex = join('|', self::$constantsForExtraction);
+        $constantsForRegex = join('|', array_merge(self::$wpConstantsForExtraction, $vpPublicConstants));
```
JanVoracek: We definitely don't want to move all public constants to the common config. For example `VP_ENVIRONMENT` or `VP_GIT_BINARY` are environment-specific. Thus, they need to stay in the `wp-config.php`.

JanVoracek:
I'm still not sure about using `constant` function to get the value of the constant. I need some time to think about it.

octopuss:
Ok so i will revert it and we will use just list of VP constants that
should be removed. I think that it will be cleaner, more straight forward
solution without big drawback.

On Wed, May 11, 2016 at 4:18 PM, Jan Voráček <notifications@github.com>
wrote:

> I'm still not sure about using constant function to get the value of the
> constant. I need some time to think about it.
>
> —
> You are receiving this because you were assigned.
> Reply to this email directly or view it on GitHub
> <https://github.com/versionpress/versionpress/pull/1042#issuecomment-218473333>
>



-- 
S pozdravom a pekný deň praje,
Ing. Ivan Doležal
http://www.octopuss.sk

plugins/versionpress/src/Utils/WpConfigEditor.php
```diff
@@ -38,6 +45,20 @@ public function updateConfigConstant($constantName, $value, $usePlainValue = fal
...
 
     /**
+     * Removes VersionPress public constants from config files.
+     * @param array $configFiles List of config file paths from which constants should be removed.
+     */
+    public static function removeVersionPressConstants($configFiles)
+    {
+        foreach ($configFiles as $configFile) {
+            $constantsForRegex = join('|', self::$vpPublicConstants);
+            $defineRegexPattern = "/(define\\s*\\(\\s*['\"]($constantsForRegex)['\"]\\s*,.*\\)\\s*;)/m";
```
JanVoracek: Can you please add a link to regex101?

plugins/versionpress/setup-hooks.php
```diff
@@ -659,6 +659,11 @@ function vp_admin_post_confirm_deactivation()
         FileSystem::remove(VERSIONPRESS_ACTIVATION_FILE);
     }
 
+    $wpConfigPath = \VersionPress\Utils\WordPressMissingFunctions::getWpConfigPath();
```
JanVoracek: This should be in the uninstallation script. We don't want to remove the constants when a user only deactivates VersionPress and activates it again (for example during an update).

JanVoracek:
It was merged into wrong branch (`master` instead of `3.x`). I re-merged it in 17bef5bd1f0cefe41191e476906d85dad61b064e.

PR #1050, opened 2016-05-19
Make AJAX work with PHP error / notice output

### LINKED ISSUE #984: Make AJAX work with PHP error / notice output
"GUI not loading" is one of the most common support requests we get, and it's usually caused by some PHP notice interfering with the REST API output (JSON parsing fails on the client).

We should improve the situation and load the UI even if there are some notices or other non-JSON output in the response. There are two levels on where we could improve things:

1. On the server side. There are [tricks][1] with output buffering available.
2. On the client side. This is probably where the focus should be because in the end, client must deal with the malformed response.

  [1]: http://wordpress.stackexchange.com/a/184238/12248

vasek17:
Resolves #984.

The API now adds error output to the response. Frontend app tries to find JSON object containing '__VP__' regardless of non-JSON output in the response.

Reviewers:

- [x] @borekb 
- [ ] @JanVoracek 
- [x] @octopuss


vasek17:
nytashzqoe

borekb:
:+1: from the UI perspective. I'll let the code review to others.

frontend/src/common/FlashMessage.react.tsx
```diff
@@ -16,9 +28,26 @@ export default class FlashMessage extends React.Component<FlashMessageProps, {}>
 
     return (
       <div className={this.props.code}>
-        <p>{this.props.message}</p>
+        <p>
+          {this.props.message} {' '}
+          {this.props.details
+            ? <a
+                className={'FlashMessage-detailsLink' + (this.state.showDetails ? '-displayed' : '-hidden')}
+                href=""
```
octopuss: How is empty `href` attribute rendered ? Is it valid to leave empty but I think that placing `#` there will be more conventional.

frontend/src/pages/HomePage.react.tsx
```diff
@@ -145,7 +150,13 @@ export default class HomePage extends React.Component<HomePageProps, HomePageSta
     WpApi
       .get('display-welcome-panel')
       .end((err: any, res: request.Response) => {
-        if (res.body === true) {
+        if (err) {
+          return;
+        }
+        const body = Array.isArray(res.body)
+          ? res.body[0]
+          : res.body;
```
octopuss: Why is this ternary condition on three adjacent lines ? Is it intentional or not.

frontend/src/common/FlashMessage.react.tsx
vasek17: The `href` parameter should be present (so it behaves like a link) but the value is optional (see [this line](https://github.com/versionpress/versionpress/blob/ac952264da8e22ec43caff77ddce1aea968a5e82/frontend/src/common/FlashMessage.react.tsx#L49))

octopuss: Ok I understand that it should be there, but I haven't seen it empty many times. Usually those cases contain `#` or `javascript:;`.

frontend/src/common/StringUtils.ts
```diff
@@ -52,3 +52,55 @@ export function pluralize(word: string) {
 export function join(array: any[], separator: string = ', ' , lastSeparator: string = ' and ') {
   return ArrayUtils.interspace(array, separator, lastSeparator).join('');
 }
+
+export function getValidVPJSON(str: string) {
```
JanVoracek: Do you have some tests for this function? It looks quite error prone.

frontend/src/pages/HomePage.react.tsx
vasek17: Fixed

frontend/src/common/FlashMessage.react.tsx
vasek17: OK, I fixed it to be consistent with the rest of code

frontend/src/common/StringUtils.ts
vasek17: Added in fa76cb2

vasek17:
Wrong base branch - see #1055.

PR #1053, opened 2016-05-19
Prevent caching of REST API requests

### LINKED ISSUE #1052: Prevent caching of REST API requests
Some web servers may cache requests made to VP REST API endpoints. It can introduce strange behaviour like https://github.com/versionpress/support/issues/46.
The most simple solution probably is to add timestamp to the query for each request.

vasek17:
Resolves #1052.

Reviewers:

- [x] @JanVoracek

### DIFF
--- a/frontend/src/services/WpApi.ts
+++ b/frontend/src/services/WpApi.ts
@@ -5,12 +5,7 @@ import config from '../config';
 
 const noCache = function (request) {
   var timestamp = Date.now().toString();
-  if (request._query !== undefined && request._query[0]) {
-    request._query[0] += '&' + timestamp;
-  } else {
-    request._query = [timestamp];
-  }
-
+  request.query(timestamp);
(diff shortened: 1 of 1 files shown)


frontend/src/services/WpApi.ts
```diff
@@ -3,6 +3,17 @@
 import * as request from 'superagent';
 import config from '../config';
 
+const noCache = function (request) {
+  var timestamp = Date.now().toString();
+  if (request._query !== undefined && request._query[0]) {
+    request._query[0] += '&' + timestamp;
```
borekb: I don't know the API but this feels a bit hacky :) Cannot `timestamp` be just added to the `_query` array?

vasek17: 👍

vasek17:
Wrong base branch - see #1054.

PR #1056, opened 2016-05-19
Adds "dirpath" to "wp vp clone"

### LINKED ISSUE #1063: Allow cloning into subdirectory
Currently, `wp vp clone` command creates a sibling directory. There's [some demand](https://github.com/versionpress/support/issues/49) to also support nested directories and we already got PR https://github.com/versionpress/versionpress/pull/1056 for that from @unclhos. However, I'm not sure we want to support that, for example, will repo inside a repo cause any issues? Won't it be confusing for the users? Create this issue to discuss that.

unclhos:
Resolves #1063

"wp vp clone" assumes going to sibling path. Add parameter to allow relative path.

I can't test this as I don't have php5.6+ locally yet. But code looks correct

Reviewers:
- [ ] @octopuss 
- [ ] @janvoracek


borekb:
Created issue #1063 to discuss whether we should or should not add this.

PR #1060, opened 2016-05-21
More reliable implementation of `WpAutomation::ensureCleanInstallationIsAvailable()`

### LINKED ISSUE #1059: More reliable `WpAutomation::ensureCleanInstallationIsAvailable()` caching
WpAutomation [caches](https://github.com/versionpress/versionpress/blob/bd898332523ec68097315f3de7cf2e385537d208/plugins/versionpress/tests/Automation/WpAutomation.php#L557-L557) clean WordPress installations to make setting up test sites faster. However, if there is an internet connection problem and the WP installation folder stays empty, WpAutomation is forever broken until a developer manually goes to `~/.wp-cli/cache/clean-installations` and removes the invalid folder like `4.5.0`.

The `WpAutomation::ensureCleanInstallationIsAvailable()` should do some sanity checks that the downloaded installation seems roughly fine. Note: we probably won't be able to use `wp core verify-checksums` as that can take quite a long time (over 10 seconds) and it would not be practical.

borekb:
Resolves #1059

Added basic check that `<cached-WP-installation>/wp-settings.php` exists. If not, will re-download the WP version from wordpress.org.

Reviewers:

- [x] @JanVoracek

### DIFF
--- a/plugins/versionpress/tests/Automation/WpAutomation.php
+++ b/plugins/versionpress/tests/Automation/WpAutomation.php
@@ -557,12 +557,17 @@ class WpAutomation
     private function ensureCleanInstallationIsAvailable()
     {
 
-        if (!is_dir($this->getCleanInstallationPath())) {
-            $downloadPath = $this->getCleanInstallationPath();
-            FileSystem::mkdir($downloadPath);
+        $cleanInstallationPath = $this->getCleanInstallationPath();
+
+        if (!is_dir($cleanInstallationPath) || !is_file($cleanInstallationPath . '/wp-settings.php')) {
+            if (!is_dir($cleanInstallationPath)) {
+                FileSystem::mkdir($cleanInstallationPath);
+            } else {
+                FileSystem::removeContent($cleanInstallationPath);
+            }
             $wpVersion = $this->siteConfig->wpVersion;
             $wpLocale = $this->siteConfig->wpLocale;
-            $downloadCommand = "wp core download --path=\"$downloadPath\" --version=\"$wpVersion\"";
+            $downloadCommand = "wp core download --path=\"$cleanInstallationPath\" --version=\"$wpVersion\"";
             if ($wpLocale) {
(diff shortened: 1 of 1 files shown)


plugins/versionpress/tests/Automation/WpAutomation.php
```diff
@@ -557,12 +557,17 @@ private function prepareFiles()
     private function ensureCleanInstallationIsAvailable()
     {
 
-        if (!is_dir($this->getCleanInstallationPath())) {
-            $downloadPath = $this->getCleanInstallationPath();
-            FileSystem::mkdir($downloadPath);
+        $cleanInstallationPath = $this->getCleanInstallationPath();
+
+        if (!is_dir($cleanInstallationPath) || !is_file($cleanInstallationPath . '/wp-settings.php')) {
+            if (!is_dir($cleanInstallationPath)) {
```
borekb: These conditions seem a bit ugly to me, if you have a better idea of how to implement them feel free to, @JanVoracek.

JanVoracek:
```php
if (!is_dir($cleanInstallationPath) || !is_file($cleanInstallationPath . '/wp-settings.php')) {
    FileSystem::remove($cleanInstallationPath);
    FileSystem::mkdir($cleanInstallationPath);

    $wpVersion = $this->siteConfig->wpVersion;
    ...
```

What about this? It looks more straightforward to me. Maybe we could also extract `is_file($cleanInstallationPath . '/wp-settings.php')` into method like `isCorrectlyDownloaded` or something like that with a comment about this good-enough solution.

borekb: Yes, that looks better. I also agree with extracting the check method. Will do that.

borekb: Alright, done, it's ready for review @JanVoracek.

PR #1065, opened 2016-05-25
Plugins, themes and WordPress installed via Composer

### LINKED ISSUE #1037: Plugins, themes and WordPress installed via Composer
Compatibility with Composer is quite important for VersionPress. We use it to declare and install the dependencies since we decided to drop the compatibility with PHP 5.2. Now it's time to move forward and utilize Composer even more.

VersionPress should support plugins, themes or even WordPress installed via Composer. This includes :
 - installation, 
 - uninstallation, 
 - updates, 
 - undo / rollback etc.

Generally, VersionPress will work with `composer.json` and `composer.lock` and run according Composer commands. Also, `composer require`, `composer update` etc. could run some WP-CLI command to commit the changes.

I've created a Composer-based WordPress stack (codename [Pedestal](https://github.com/versionpress/pedestal)) inspired by Bedrock. We can utilize it to test the workflows with Composer.

JanVoracek:
Resolves #1037.

VersionPress now works with Composer!

### Tracking chages in `composer.json`
VersionPress now tracks installation / uninstallation / update of Composer packages using [Composer scripts](https://getcomposer.org/doc/articles/scripts.md). There are two scripts:
- `pre-update-cmd` calls `wp vp-composer prepare-for-composer-changes`. This command saves list of plugins and themes installed before Composer changes files into transient option.
- `post-update-cmd`calls `wp vp-composer commit-composer-changes`. This command creates according `ChangeInfo` objects and does the commit.

### Installing correct dependencies after VP actions
VersionPress now calls `composer install` after:
- undo / rollback
- clone
- push / pull
- restore-site

### Support of Composer-based site in `test-config.yml`
`SiteConfig` now contains `type` to define standard / composer-based site.  `WpAutomation::runAutomation` then installs standard WP site using WP-CLI or [Pedestal-based](https://github.com/versionpress/pedestal) site using `composer create-project`.

### Several fixes in tests
It was necessary to adjust the tests to work with this different approach. For example, there were often hardcoded paths like `wp-content/vpdb`, `wp-includes` etc.

Reviewers:
- [x] @octopuss 
- [ ] @borekb

### DIFF
--- a/plugins/versionpress/src/Cli/vp-composer.php
+++ b/plugins/versionpress/src/Cli/vp-composer.php
@@ -1,0 +1,121 @@
+<?php
+// NOTE: VersionPress must be fully activated for these commands to be available
+
+// WORD-WRAPPING of the doc comments: 75 chars for option description, 90 chars for everything else,
+// see http://wp-cli.org/docs/commands-cookbook/#longdesc.
+// In this source file, wrap long desc at col 97 and option desc at col 84.
+
+namespace VersionPress\Cli;
+
+use VersionPress\ChangeInfos\PluginChangeInfo;
+use VersionPress\DI\VersionPressServices;
+use VersionPress\Git\Committer;
+use VersionPress\Utils\Process;
+use VersionPress\VersionPress;
+use WP_CLI;
+use WP_CLI_Command;
+
+/**
+ * VersionPress CLI commands for Composer scripts.
+ */
+class VPComposerCommand extends WP_CLI_Command
+{
+    /**
+     * Commits all changes made by Composer.
+     *
+     * @subcommand commit-composer-changes
+     */
+    public function commitComposerChanges($args, $assoc_args)
+    {
+        global $versionPressContainer;
+
+        if (!VersionPress::isActive()) {
+            WP_CLI::error('VersionPress is not active. Changes will be not committed.');
+        }
+
+        /** @var Committer $committer */
+        $committer = $versionPressContainer->resolve(VersionPressServices::COMMITTER);
+        $changes = $this->detectChanges();
+        $installedPackages = $changes['installed'];
+        $removedPackages = $changes['removed'];
+        $updatedPackages = $changes['updated'];
+
+        $changeInfos = array_merge(
+            array_map(function ($package) {
+                if ($package['type'] === 'wordpress-plugin') {
+                    return new PluginChangeInfo($package['name'], 'install', $package['name']);
+                }
+
+                return null;
+
+            }, $installedPackages),
+            array_map(function ($package) {
+                if ($package['type'] === 'wordpress-plugin') {
+                    return new PluginChangeInfo($package['name'], 'delete', $package['name']);
+                }
+
+                return null;
+
+            }, $removedPackages),
+            array_map(function ($package) {
+                if ($package['type'] === 'wordpress-plugin') {
+                    return new PluginChangeInfo($package['name'], 'update', $package['name']);
+                }
+
+                return null;
+
+            }, $updatedPackages)
+        );
+
+        var_dump($changeInfos);
+    }
+
+    private function detectChanges()
+    {
+        $currentComposerLock = file_get_contents(VP_PROJECT_ROOT . '/composer.lock');
+
+        $process = new Process(VP_GIT_BINARY . ' show HEAD:composer.lock', VP_PROJECT_ROOT);
+        $process->run();
+
+        $previousComposerLock = $process->getOutput();
+
+        $currentPackages = $this->getPackagesFromLockFile($currentComposerLock);
+        $previousPackages = $this->getPackagesFromLockFile($previousComposerLock);
+
+        $installedPackages = array_diff_key($currentPackages, $previousPackages);
+        $removedPackages = array_diff_key($previousPackages, $currentPackages);
+
+        $packagesWithChangedVersion = array_filter(
+            array_intersect_key($previousPackages, $currentPackages),
+            function ($package) use ($currentPackages) {
+                return $package['version'] !== $currentPackages[$package['name']]['version'];
+            }
+        );
+
+        return [
+            'installed' => $installedPackages,
+            'removed' => $removedPackages,
+            'updated' => $packagesWithChangedVersion,
+        ];
+    }
+
+    private function getPackagesFromLockFile($lockFileContent)
+    {
+        $lockFile = json_decode($lockFileContent, true);
+        return array_combine(
+            array_column($lockFile['packages'], 'name'),
+            array_map(function ($package) {
+                return [
+                    'name' => $package['name'],
+                    'version' => $package['version'],
+                    'type' => $package['type'],
+                    'homepage' => @$package['homepage'] ?: null,
+                ];
+            }, $lockFile['packages'])
+        );
+    }
+}
+
+if (defined('WP_CLI') && WP_CLI) {
+    WP_CLI::add_command('vp-composer', VPComposerCommand::class);
+}
--- a/plugins/versionpress/src/Cli/vp-composer.php
+++ b/plugins/versionpress/src/Cli/vp-composer.php
@@ -7,6 +7,7 @@
 
 namespace VersionPress\Cli;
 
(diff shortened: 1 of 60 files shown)


plugins/versionpress/tests/Automation/WpAutomationRunner.sample.php
```diff
@@ -28,6 +30,46 @@ public function runAutomation()
     /**
```
octopuss: Please add link to `versionpress/pedestal` repository. Users/developers may not be familiar with it.

plugins/versionpress/src/Cli/vp-composer.php
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+        /** @var Committer $committer */
+        $committer = $versionPressContainer->resolve(VersionPressServices::COMMITTER);
+        $changes = $this->detectChanges();
+        $installedPackages = $changes['installed'];
+        $removedPackages = $changes['removed'];
+        $updatedPackages = $changes['updated'];
+
+        $changeInfos = array_merge(
```
octopuss: These code duplicities may be replaced with stg. like this https://3v4l.org/vXH4a

plugins/versionpress/src/Cli/vp-composer.php
```diff
@@ -20,6 +22,17 @@
  */
 class VPComposerCommand extends WP_CLI_Command
 {
+
+    private $pluginsThemesTransient = 'vp_composer_plugins_themes';
```
octopuss: is `Transient` a term for naming transient options in WordPress world, or there can be also other transient objects/values/etc.

My note is if the variable is named correctly.

plugins/versionpress/src/Cli/vp-composer.php
JanVoracek: This is an outdated code and now it looks like [this](https://github.com/versionpress/versionpress/pull/1065/files#diff-9d8b13eddfd1eed01669006028cf0a0cR65). Are you sure you are reviewing the whole PR?

octopuss: Heh, I started to review it commit by commit for once and now I realized that it was not a good idea. You've refactored it a few commits later so concern this as outdated comment :)

plugins/versionpress/src/Utils/RequirementsChecker.php
```diff
@@ -121,6 +120,16 @@ public function __construct($database, DbSchemaInfo $schema, $checkScope = Requi
             'help' => 'VersionPress automatically tries to secure certain locations, like `wp-content/vpdb`. You either don\'t have a supported web server or rules cannot be enforced. [Learn more](http://docs.versionpress.net/en/getting-started/installation-uninstallation#supported-web-servers).'
         ];
         if ($checkScope === RequirementsChecker::SITE) {
+            if (file_exists(VP_PROJECT_ROOT . '/composer.json')) {
+                $this->requirements[] = [
+                    'name' => 'Composer scripts',
+                    'level' => 'warning',
+                    'fulfilled' => $this->testComposerJson(),
+                    // @codingStandardsIgnoreLine
+                    'help' => 'Your project uses Composer but `pre-update-cmd` or `post-update-cmd` scripts in `composer.json` are already defined. VersionPress needs to run a WP-CLI commands in these scripts to automatically commit changes of Composer packages. Please update your scripts to run [the WP-CLI commands](http://docs.versionpress.net/en/feature-focus/composer).'
```
octopuss: Please create issue in documentation to create content for `feature-focus/composer`.

plugins/versionpress/src/Cli/vp-composer.php
```diff
@@ -0,0 +1,179 @@
...
+
+            if ($packageName === $pluginDirectory) {
+                return new PluginChangeInfo($pluginFile, $action, $plugin['Name']);
+            }
+        }
+
+        return null;
+    }
+
+    private function createThemeChangeInfo($action, $fullPackageName, $themes)
```
octopuss: Shouldn't this method be placed to some kind of factory, or as static to `ThameChangeInfo.php` ? I have same question for the method `createPluginChangeInfo`.

plugins/versionpress/src/Initialization/Initializer.php
```diff
@@ -635,4 +637,27 @@ private function getEntitiesFromDatabase($entityName)
...
+
+        if (!isset($composerJson->scripts)) {
+            $composerJson->scripts = new \stdClass();
+        }
+
+        if ((!isset($composerJson->scripts->{'pre-update-cmd'}) || $composerJson->scripts->{'pre-update-cmd'} === '') &&
+            (!isset($composerJson->scripts->{'post-update-cmd'}) || $composerJson->scripts->{'post-update-cmd'} === '')
+        ) {
+            $composerJson->scripts->{'pre-update-cmd'} = 'wp vp-composer prepare-for-composer-changes';
+            $composerJson->scripts->{'post-update-cmd'} = 'wp vp-composer commit-composer-changes';
```
octopuss: What if user does not have `wp-cli` installed ? how will `pre-update-cmd` and `post-update-cmd` commands behave ?

plugins/versionpress/src/Cli/vp-composer.php
JanVoracek: Yes, it is – https://codex.wordpress.org/Transients_API

plugins/versionpress/src/Utils/RequirementsChecker.php
JanVoracek: Done https://github.com/versionpress/docs/issues/59

plugins/versionpress/src/Initialization/Initializer.php
JanVoracek: Like that
```zsh
➜  vp01 git:(master) composer remove wpackagist-plugin/akismet
> wp vp-composer prepare-for-composer-changes
sh: wp: command not found
Script wp vp-composer prepare-for-composer-changes handling the pre-update-cmd event returned with an error

Removal failed, reverting ./composer.json to its original content.


  [RuntimeException]
  Error Output: sh: wp: command not found
```

plugins/versionpress/src/Cli/vp-composer.php
JanVoracek: I think creating a factory for two short methods is an overkill. Plus, they are closely bound to Composer; hence, I think they have every right to be here.

plugins/versionpress/tests/Automation/WpAutomationRunner.sample.php
JanVoracek: Done [here](https://github.com/versionpress/versionpress/blob/083002bc337ca2c07a0868df2a338e23b3c0f202/plugins/versionpress/tests/Automation/WpAutomation.php#L828-L832).

plugins/versionpress/src/Cli/vp-composer.php
octopuss: Factory was an option, static method in a class was another one. I agree that factory may be too much. If I look at it as on `wp-cli` command, link to `composer` is not clear . To me, it is similar case as `buildFromCommitMessage` method which is also static.

plugins/versionpress/src/Initialization/Initializer.php
octopuss: So it is ok.

plugins/versionpress/src/Cli/vp-composer.php
JanVoracek: The parameters of this methods are `$fullPackageName` (name of a Composer package, e.g. `wpackagist-plugin/akismet`) and list of plugins installed in WordPress. This method only finds the plugin in the list that matches with the name of Composer package.

plugins/versionpress/src/Initialization/Initializer.php
JanVoracek: Maybe we should check WP-CLI in the Requirements Checker?

octopuss: Maybe as a soft requirement/recommendation when `composer.json` is found in the client's project.

PR #1070, opened 2016-05-30
Make AJAX work with PHP error / notice output

### LINKED ISSUE #984: Make AJAX work with PHP error / notice output

vasek17:
Resolves <del>#984</del> #1073.

Reviewers:

- [ ] @JanVoracek

### DIFF
--- a/frontend/src/pages/HomePage.react.tsx
+++ b/frontend/src/pages/HomePage.react.tsx
@@ -230,7 +230,7 @@ export default class HomePage extends React.Component<HomePageProps, HomePageSta
           if (err) {
             reject(HomePage.getErrorMessage(res, err));
           } else {
-            resolve(res.body);
+            resolve(res.body.status);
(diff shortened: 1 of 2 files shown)


borekb:
We've already shipped 3.0.1 so this cannot resolve #984. I've updated the issue number but it will be better to fix the problem by implementing a better response handling in general (#1074) anyway so I'm closing this.

PR #1075, opened 2016-06-01
Plugin support

### LINKED ISSUE #1036: Plugin support, 4.0-alpha1 version
Support for 3rd party plugins (and themes, to a lesser degree) is a major theme of v4.0. VersionPress needs to understand plugins' data and actions in order to provide version control for them.

Previous releases introduced things like [ChangeInfo objects](https://github.com/versionpress/versionpress/blob/622a2fca06a71e6e8ad802901b84fa16ed4563be/plugins/versionpress/src/ChangeInfos/ChangeInfo.php), [schema format](https://github.com/versionpress/versionpress/blob/622a2fca06a71e6e8ad802901b84fa16ed4563be/plugins/versionpress/src/Database/wordpress-schema.yml) ([readme](https://github.com/versionpress/versionpress/blob/622a2fca06a71e6e8ad802901b84fa16ed4563be/plugins/versionpress/src/Database/schema-readme.md)) etc. which we're going to build upon and make more extensible. There will probably be many smaller, concrete tickets but in general, we should:

1. Support loading [DbSchemaInfo](https://github.com/versionpress/versionpress/blob/622a2fca06a71e6e8ad802901b84fa16ed4563be/plugins/versionpress/src/Database/DbSchemaInfo.php) from **multiple schema-files**. For example, a WooCommerce site would load `wordpress-schema.yml` and `woocommerce-schema.yml`.
    - Similarly with [ShortcodesInfo](https://github.com/versionpress/versionpress/blob/622a2fca06a71e6e8ad802901b84fa16ed4563be/plugins/versionpress/src/Database/ShortcodesInfo.php) and [`-shortcodes.yml`](https://github.com/versionpress/versionpress/blob/622a2fca06a71e6e8ad802901b84fa16ed4563be/plugins/versionpress/src/Database/wordpress-shortcodes.yml) files.
2. Develop a **discovery mechanism for schema-files**. The idea is to look in three locations: VersionPress itself, plugin's source code (over time, schemata will hopefully ship with plugins if VersionPress becomes popular enough) or a central repository, possibly backed by a GitHub repo so that everyone can send a pull request.
3. **Reconsider ChangeInfo objects** structure and possibly simplify it (#1034). ChangeInfos (or equivalent) should also be shipped similarly to schema-files.
4. Work out a way to **describe actions**. We don't have a great abstraction for this yet, for example, actions are currently only described in PhpDoc sections of ChangeInfo files, cannot be processed automatically, etc. We already have technical tickets like #804 to improve this.

---

🚧 Current status: VersionPress' code is being refactored in **PR https://github.com/versionpress/versionpress/pull/1075**. Current status can be tracked in ➡️ https://github.com/versionpress/versionpress/pull/1075#issuecomment-239466089.

---

☝️  This ticket is a good place to solicit feedback from plugin authors. If you're one of them and would be interested in integrating with VersionPress, please let us know in the comments below. So far, we've heard interest from these plugins / folks:

- [Toolset / Types](https://wp-types.com/) – @zaantar
- [Pods Framework](http://pods.io/) – @sc0ttkclark
- [WooCommerce](https://www.woothemes.com/woocommerce/) – @pavelevap, @conschneider

---

Preliminary list of **things we'll likely need to improve in VersionPress** to support certain plugins:

- Dynamic table schemas – https://github.com/versionpress/versionpress/issues/1036#issuecomment-217214337
- Dynamic targets for standard references – https://github.com/versionpress/versionpress/issues/1036#issuecomment-217389148
- Ability to specify paths to ignore – https://github.com/versionpress/versionpress/issues/1036#issuecomment-219499106

borekb:
Resolves #1036.

Opening this PR early, it will be a long-running one, probably.


JanVoracek:
Closed by mistake; reopening.

JanVoracek:
## Refactoring
- Storages:
  - [x] Replace classes specific for entities with generic ones
  - [x] Extract code from `shouldBeSaved` methods to filters
  - [x] Extract logic for creating ChangeInfo objects to filters
- EntityInfo:
  - [x] Replace classes specific for entities with generic ones
  - [x] Extract logic for the detection of an action to filters
  - [x] Implement automatic filling of tags from entities
  - [x] Add filter for custom tags
  - [x] Add filter for custom paths
  - [x] Introduce priorities for actions and sort the ChangeInfo objects by them
  - [x] Support using values of tags in commit messages
  - [x] Add filter for customizing the commit message
  - [x] Replace `Committer::forceChangeInfo` with a simple function
- Synchronizers
  - [ ] Replace classes specific for entities with generic ones
  - [ ] Move code from `doEntitySpecificActions` in filters
  - [ ] Support custom actions before and after sync
  - [ ] Introduce better syntax for cache invalidation
  - [ ] Replace `transformEntities` with a filter
- GUI/API
  - [ ] Introduce some markup language for formatting of commit messages
  - [ ] Update GUI to support the markup language 
- Committer
  - [ ] Replace pre-processing mechanism of ChangeInfo classes with WP filters (we need some simple format instead of ChangeInfo here)
  - [ ] Use multiple sources for `ChangeInfo` objects instead of `forceChangeInfo` method

borekb:
Here are some notes as I reviewed `Plugin-Support.md`:

- [ ] It would be good to have a clear name for the set of files in the `.versionpress` folder. "Plugin definition" is the best I can currently think of.
- [ ] We have been using the `edit` action since VersionPress 1.0 (e.g., `option/edit`), however, the more natural English verb is `update`. 4.0 would be a chance to fix this. (Feedback from native speakers welcome.)
- [ ] Do we need the separate `_meta_` entity filters, e.g., `vp_entity_tags_` vs. `vp_meta_entity_tags_`? I get that they are slightly different, however, if a single filter name could be used for both cases, it would make the API smaller. Let's discuss this, @JanVoracek.
- [ ] Is "schema" the right word? Our `schema.yml` contains quite a bit more. On the other hand, "schema" is quite unique and intuitively connected to a database so it might be OK to call it that.
- [ ] `.versionpress` makes sense in the plugin folder but for the planned online repo, we need a convention to separate the plugins. See e.g. [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped).
- [ ] Our term "actions" conflicts with WP actions. Not be a big deal for us but WP developers unfamiliar with VersionPress might be confused. If needs be, we can consider to use "events" instead.
    - > As nouns the difference between action and event is that action is something done so as to accomplish a purpose while event is an occurrence; something that happens. -- http://wikidiff.com/event/action 
- [ ] Document the slash in Actions

plugins/versionpress/src/ChangeInfos/UntrackedChangeInfo.php
```diff
@@ -26,13 +26,13 @@ public function getCommitMessage()
...
     {
-        return new self($commitMessage);
+        return $this->commitMessage->getUnprefixedSubject();
     }
 
-    public function getChangeDescription()
+    public function getPriority()
     {
-        return $this->commitMessage->getUnprefixedSubject();
+        return 0;
```
octopuss: When no better  is found, why priority is not high instead ?

JanVoracek: `UntrackedChangeInfo` represents manual commit; thus, there are no other `ChangeInfo` objects to compare with. Plus, ignoring the fact it will not compare with anything, zero is the highest priority.

docs/Plugin-Support.md
```diff
@@ -0,0 +1,583 @@
...
+
+Plugins are described to VersionPress by a set of files stored in the `.versionpress` folder in the plugin root (other discovery options are described below). They include:
+
+- Actions in `actions.yml`
+- Database schema in `schema.yml`
+- Shortcodes in `shortcodes.yml`
+- Hooks are defined in `hooks.php`
+
+All files are optional so for example, if a plugin doesn't define any new shortcodes it can omit the `shortcodes.yml` file. Simple plugins like _Hello Dolly_ might even omit everything; they just need to have the `.versionpress` folder so that VersionPress knows the plugin is supported.
+
```
JanVoracek: I don't really know. Originally, I didn't intend to use this directory to let VP know if the plugin is supported. It's maybe question to @borekb.

docs/Plugin-Support.md
```diff
@@ -0,0 +1,583 @@
...
+  ['type' => 'all-storage-files', 'entity' => 'option'],
+  ['type' => 'path', 'path' => '/var/www/wp/example.txt'],
+  ['type' => 'path', 'path' => '/var/www/wp/folder/*']
+]
+```
+
+### Non-database actions
+
+Some actions are not directly related to the database entities, e.g. plugin installation, WP update, etc. VersionPress provides function `vp_force_action` for these actions. VersionPress will use only action specified by parameters of this function and ignore all automatically catched. For example:
+
```
JanVoracek: Not only the priority – the actions are ignored entirely. Do you have any suggestion how to reword the sentence?

docs/Plugin-Support.md
```diff
@@ -0,0 +1,583 @@
...
+> Warning: This function is only a temporary solution and will be removed in 4.0.
+
+## Database schema
+
+If the plugin adds custom data into the database it must provide a `schema.yml` file describing the database model. For example, this is how WordPress posts are described:
+
+```yaml
+post:
+  table: posts
+  id: ID
```
JanVoracek: VersionPress doesn't support it yet.

docs/Plugin-Support.md
borekb: `.gitkeep` is an option but it's true that we didn't discuss this yet. The empty `.versionpress` folder seemed to me like the easiest way for a plugin author to declare explicit support for VersionPress even if no definitions are necessary.

Or maybe it could be a `.versionpress` file?

docs/Plugin-Support.md
borekb: I wanted to look at this section anyway ("catched" etc.), will try to reword it.

docs/Plugin-Support.md
JanVoracek: I think it should be some service. The main reason is that you don't have the files before you install the plugin; therefore, you cannot say whether the plugin is supported or not on the `/plugin-install.php` page.

PR #1079, opened 2016-06-15
Improve invalid REST API response handling

### LINKED ISSUE #1074: Improve invalid REST API response handling
In #984, we've worked around the situation where the server (REST API) sends a malformed response, e.g., by including a PHP notice from some plugin. The `__VP__` key has been added to every JSON response so that the client-side JavaScript code can understand what is the real response. However, after some internal discussion, we're not sure this is the right approach.

This issue is about reconsidering our options here. Some initial comments:

- The current approach (as per #984) modifies the object and feels slightly weird. We'll probably abandon it.
- Every response could be wrapped in an object like this:

    ```
    {
      __VP__: true,
      data: theActualValue
    }
    ```

    The client would use something like `response.body.data.theActualValue` instead of `response.body.theActualValue`. This would also provide the opportunity to send additional metadata with every response should we need that, like what VersionPress or WP version are running on the server.
- The client could also reject to work if the response is malformed. I think we considered this #984 but in the end decided it is better for the users if the UI works despite some warnings in it. However, it could be argued that sending notices and such to the client is a security risk / server misconfiguration and we could just show some good error to guide the users to fix their server setup. This is a valid approach.

Let's do some research because the other projects relying on WP REST API must hit the same issue as well.

vasek17:
Resolves #1074.

Updated API response format to send the actual value in the `data` property instead of modifying the value itself. Also added type definitions for API responses.

Reviewers:

- [x] @borekb 
- [x] @octopuss


frontend/src/services/Wp.d.ts
```diff
@@ -0,0 +1,25 @@
+declare module Wp {
```
borekb: Perhaps the better name would be `VpApi`? Seems very specific to VersionPress and its API.

vasek17: 👍

octopuss:
@vasek17 : I see you've renamed `Wp` to `VpApi` so it is ok. Please do so for `Wp.d.ts` file as well.

JanVoracek:
@borekb, @octopuss: Do you have any other comments?

borekb:
@JanVoracek No it seems fine.

PR #1096, opened 2016-07-12
Improve unclear attribute names

### LINKED ISSUE #1085: Fix inconsistent and unclear naming convention of attributes

The front end codebase contains inconsistent and unclear attribute names.

For example, in the BulkActionPanel component there is a property **selected**. When you first encounter name of the property you think that the author probably meant a bool value. But a further examination of the component shows that it is actually an array of **Commit**s. Easy fix in this example would be to rename the property to **_selectedCommits_**.

Another problem are names of attributes containing a bool value. For example **loading**, **display** etc. Better names for these attributes could be **_isLoading_**, **_isVisible_** etc. But in this case I am open to discussion.

Ishirak:
Resolves #1085.

Improved sometimes unclear attribute names in the front end codebase. If you have any suggestion to improve other names, feel free to share them.

Reviewers:

- [x] @JanVoracek

### DIFF
--- a/frontend/src/BulkActionPanel/BulkActionPanel.react.tsx
+++ b/frontend/src/BulkActionPanel/BulkActionPanel.react.tsx
@@ -8,7 +8,7 @@ interface BulkActionPanelProps extends React.Props<JSX.Element> {
   enableActions: boolean;
   onBulkAction: (action: string) => void;
   onClearSelection: () => void;
-  selected: Commit[];
+  selectedCommits: Commit[];
 }
 
 export default class BulkActionPanel extends React.Component<BulkActionPanelProps, {}> {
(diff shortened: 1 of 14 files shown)


JanVoracek:
`node ./config/webpack-build.js` fails on:

```
Error: ./src/pages/HomePage.react.tsx
(190,13): error TS2345: Argument of type '{ displayUpdateNotice: boolean; dirtyWorkingDirectory: boolean; }' is not assignable to parameter of type 'HomePageState'.
  Object literal may only specify known properties, and 'dirtyWorkingDirectory' does not exist in type 'HomePageState'.
```

JanVoracek:
I removed @vasek17 from reviewers because he's on vacation.

Issue #1088, opened 2016-07-05
Fix inconsistent order of the React.js component’s (life cycle) methods


Ishirak:
This issue is part of a frontend code review, see #1083.

Several orders are used in the front end codebase. In some components the **render** method is the last one in the order, in the others is not. For clarity and readability of the codebase is preferred to use a consistent order. 

Most common order is with **render** method as the _last one_ in the component. An example of a consistent order can be found in the [Airbnb React/JSX Style Guide](https://github.com/airbnb/javascript/tree/master/react#ordering).

vasek17:
OK, let's go with [Airbnb React/JSX Style Guide](https://github.com/airbnb/javascript/tree/master/react#ordering).

borekb:
:ship: Shipped in [4.0-alpha1](https://github.com/versionpress/versionpress/releases/tag/4.0-alpha1).

PR #1098, opened 2016-07-25
Use consistent function binding and order of React.js methods

### LINKED ISSUE #1090: Fix inconsistent techniques of binding JavaScript functions

Three binding techniques are used in the front end codebase.
- In the constructor.
- In the render method.
- Not binding at all – functions are in component's private attributes.

Binding in the constructor is a valid option and is used by many developers. 
Binding in the render method is also used by many developers, but is not a performant option – a bind call in the render method creates a brand new function on every single render. ([MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/bind))

My preffered solution is to use [arrow functions] (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions). They don't require boilerplate in the constructor, stylistically aren't much different and automatically bind themselves.

For example, constructor binding looks like this:
```JSX
    // ...
    constructor() {
        // ...
        this.onClearSelection = this.onClearSelection.bind(this);
        // ...
    }

    onClearSelection(e: React.MouseEvent) {
        // ...
    }
    // ...
```
On the other hand, arrow functions look like this:
```JSX
    // ...
    onClearSelection = (e: React.MouseEvent) => {
        // ...
    }
    // ...
```

Ishirak:
Resolves #1090 and #1088.

Fixed inconsistent JavaScript function binding by introducing [arrow functions] (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions). In the process of fixing function binding also introduced [class instance fields](https://github.com/jeffmo/es-class-fields-and-static-properties) for initializing the React.js component's state and fixed inconsistent order of the React.js component’s (life cycle) methods.

Reviewers:

- [x] @JanVoracek
- [x] @vasek17

### DIFF
--- a/frontend/src/BulkActionPanel/BulkActionPanel.react.tsx
+++ b/frontend/src/BulkActionPanel/BulkActionPanel.react.tsx
@@ -22,19 +22,19 @@ export default class BulkActionPanel extends React.Component<BulkActionPanelProp
     }
 
     this.props.onBulkAction(value);
-  }
+  };
 
   onClearSelection = (e: React.MouseEvent) => {
     e.preventDefault();
     this.props.onClearSelection();
-  }
+  };
 
   render() {
     const { selectedCommits, enableActions } = this.props;
 
     const noteClassName = classNames({
       'BulkActionPanel-note': true,
-      'hide': selectedCommits.length === 0
+      'hide': selectedCommits.length === 0,
     });
 
     return (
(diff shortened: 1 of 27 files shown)


vasek17:
I fixed some issues as I reviewed the PR. 
@Ishirak: Check the changes I made and if you are OK with them, you can merge the PR.

PR #1104, opened 2016-08-09
Refactor of BulkActionPanel component

### LINKED ISSUE #1087: Use simple and clear React.js components

Ishirak:
Resolves (partialy) #1087.

Few notes:
- Directory structure – as we agreed on – will be implemented after all of the components are refactored.

Reviewers:

- [x] @JanVoracek
- [x] @vasek17


borekb:
> I don't believe that its solution should be handled by one huge PR. Therefore I opened this PR, which will be continuously updated to resolve this issue.

So is this going to be a single large PR (though created over time) or multiple smaller PRs? It's not clear from the description.

Ishirak:
Sorry, my bad,  this going to be a single large PR (though created over time).

Ishirak:
But if anyone has in mind a different aproach, I'm open to suggestions.

borekb:
An alternative would be to create multiple smaller PRs against a single issue, as we discussed offline. The main benefit would be that smaller PRs could be merged into master sooner. But if @vasek17 doesn't depend on any of this work, single PR is fine.

Ishirak:
I've tried this alternative before I posted mine, but since it's a single issue I only have a single branch to work with. After I'd create a PR, the PR has to be reviewed and I can't work on another refactoring in the branch because it'd violate the PR with refactoring which the PR is not part of. One possible solution could be divide the issue into two branches – refactoring of codebase can be parallel – but I don't like it that much. 

Maybe I'm missing some elementary GitHub workflow and there exists an easy fix which I'm not familiar with. If so, please enlighten me.

Ishirak:
After internal discussion we are going for an alternative involving multiple smaller PRs. Thanks to @borekb for his insight on this issue.

Ishirak:
Edited PR's description.

Issue #1086, opened 2016-07-05
Fix inconsistent and unclear order of JSX attributes


Ishirak:
This issue is part of a frontend code review, see #1083.

The front end codebase contains inconsistent and unclear order of JSX attributes.

Multiple JSX tags and attributes are in some components on the same line, in other components, or even in other parts of the same component, they aren't.

For example,
```JSX
<a className='BulkActionPanel-clear' href='#' onClick={this.onClearSelection.bind(this)}>clear selection</a>
```
can be converted into
```JSX
<a 
    className="bulk-action-panel-clear"
    href="#" 
    onClick={this.onClearSelection}
>
    clear selection
</a>
```
which is much cleaner and readable.

vasek17:
In this case I would prefer (e.g. [here](https://github.com/versionpress/versionpress/blob/8778ceb405fc69cf044baf66d08fa76981c363cb/frontend/src/CommitPanel/CommitPanelCommit.react.tsx#L49))
```javascript
<a 
    className="bulk-action-panel-clear"
    href="#" 
    onClick={this.onClearSelection}
>clear selection</a>
``` 
or
```javascript
<a className="bulk-action-panel-clear" href="#" onClick={this.onClearSelection}>
    clear selection
</a>
``` 
when the text would be longer or it would be block.

I try to write the component on a single line if it's readable otherwise display it multi-line and indent.
This example is an edge case - it is ok for me both single line and multi-line.

borekb:
This is a small stylistic issue where different people will probably have different tastes (I, for example, prefer Vasek's last example). I don't think we need a hard rule here, just do whatever seems sensible and I, personally, don't have a problem reading a code where the styles are slightly mixed and matched as long as the code is readable (and Vasek's last example is perfectly fine by me).

vasek17:
@Ishirak: I don't think I get your point so maybe one more example.
If the tag has a lot of attributes it should be multi-line. If it contains block or the text is longer is could look like this:
```javascript
<a 
    className="bulk-action-panel-clear"
    href="#" 
    onClick={this.onClearSelection}
    anotherAttribute="looooooong one"
    yetAnother=true
>
    Veeeeery long lorem ipsum dolor sit amet text. ....
</a>
```

However it does not make sense to me to format small tags (extreme case) like this:
```javascript
<a 
    href="#"
>
    close
</a>
```

In this case 
```javascript 
<a href="#">close</a>
```
would be preferred.

[...]

PR #1105, opened 2016-08-11
Refactor of CommitPanel component

Ishirak:
Resolves (partialy) #1087.

Part of this refactor was introduction of `enum` type for property `detailsLevel`. I believe that `enum` is much better suited as a type for this property than `string` was – property can only contain three values. Since these values were numerously checked against in other parts of this component subtree, `enum` type provided better readability and maintainability than a "magic" string could.

If reviewers find use of `enum` type useful as I do, I'd like to discuss location of `enums.ts` file. Right now it contains just one type, but in the future that might not be the case. Since this type will be useful in other parts of #1087, e.g. [CommitsTableRow](https://github.com/versionpress/versionpress/blob/b12e17237143f1a5bcd6625f46e38da869b8f149/frontend/src/Commits/CommitsTableRow.react.tsx) component, I chose `frontend/src/enums` directory. but I'm always open to suggestions.

Reviewers:

- [x] @JanVoracek
- [x] @vasek17


vasek17:
Ad import orderings - see [HomePage](https://github.com/versionpress/versionpress/blob/b12e17237143f1a5bcd6625f46e38da869b8f149/frontend/src/pages/HomePage.react.tsx):

- React
- All other external dependencies
- <newline>
- Internal dependencies - Components
- Internal dependencies - Other
- <newline>
- Styles

Each group preferably in alphabetical order

frontend/src/commit-panel/CommitPanel.tsx
```diff
@@ -0,0 +1,114 @@
+import * as React from 'react';
+import * as classNames from 'classnames';
+
+import { DetailsLevel } from '../enums/enums';
```
vasek17: Move under the component imports.

frontend/src/commit-panel/Notice.tsx
```diff
@@ -0,0 +1,42 @@
...
+    const { detailsLevel } = this.props;
+
+    return (
+      <p>
+        {`You have `}
+        <a
+          href='http://docs.versionpress.net/en/feature-focus/undo-and-rollback#uncommitted-files'
+          target='_blank'
+        >
+          uncommitted changes
```
vasek17: I would prefer more compact style (text on the same line) when it is short (as discussed [here](https://github.com/versionpress/versionpress/issues/1086#issuecomment-231018574)).
In this case:
```jsx
<a
  href='http://docs.versionpress.net/en/feature-focus/undo-and-rollback#uncommitted-files'
  target='_blank'
>uncommitted changes</a>

File: frontend/src/commit-panel/commit/Commit.tsx [hunk 3]
```diff
@@ -0,0 +1,84 @@
...
+    e.preventDefault();
+
+    this.setState({
+      isFormVisible: true,
+    });
+  };
+
+  onCommitMessageChange = (e: React.FormEvent) => {
+    this.setState({
+      commitMessage: (e.target as any).value,
```
vasek17: `(e.target as HTMLTextAreaElement)`

Can be extracted directly in the attribute:
```
(`({ target: HTMLTextAreaElement })`)
```
 or in the body:
```
const target = e.target as HTMLTextAreaElement;
```

frontend/src/commit-panel/commit/Form.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+interface FormProps {
+  commitMessage: string;
+  onCommitMessageChange(e: React.FormEvent): void;
+  onSubmit(e: React.FormEvent): void;
+  onCancelCommitClick(e: React.MouseEvent): void;
+}
+
+const Form: React.StatelessComponent<FormProps> = (
+  { commitMessage, onCommitMessageChange, onSubmit, onCancelCommitClick }) => (
```
vasek17: I would destruct the props in the body when they are this long to avoid line breaking

frontend/src/commit-panel/Notice.tsx
Ishirak: If you **really** insist on this style, then I can make a compromise and use it. But personally I'd prefer mine, I believe that is cleaner and more readable, at least for me.

frontend/src/commit-panel/commit/Form.tsx
Ishirak: I agree that this format style is not a great one, but I don't believe that destructuring in the function body is the way to go. Recently I've used this style for such a stateless component:

```jsx
const Form: React.StatelessComponent<FormProps> = ({
  commitMessage, onCommitMessageChange, onSubmit, onCancelCommitClick,
}) => (
  // ...
);
```

What do you think about it?

frontend/src/commit-panel/Notice.tsx
borekb: Both styles are fine, let's move on.

vasek17:
Ad enums - I think it is generally good idea to use enums instead of strings. 
I'd prefer folder `frontend/src/enums/` that contains separate files for each enum (e.g. `enums/DetailsLevel.ts`, `enums/AnotherOne.ts`).

JanVoracek:
I tried to run the frontend on this branch and it seems it doesn't like me 😞 I even tried to delete `node_modules` and `npm install` from scratch. 

This is a result with some **uncommitted file** in the repository:

screen shot 2016-08-18 at 15 03 57

Any ideas?

JanVoracek:
Now it displays the uncommitted changes but I still cannot commit them. The request sent to `?rest_route=/versionpress/commit` uses `OPTIONS` HTTP method. Is it OK?

screen shot 2016-08-18 at 16 03 14

vasek17:
Yes, the **POST** requests from another domain don't work at the moment (some CORS limitations in WP REST API). Please test this request in the WP admin.

PR #1106, opened 2016-08-12
Refactor of ServicePanel and WelcomePanel components

Ishirak:
Resolves (partialy) #1087.

Since both components are smaller ones, PR contains refactor of both of them.

Because I believe in very clean `render` method of the main parent component – in this case `HomePage` – I've moved some render logic from this component to the `ServicePanel` component – VersionPress title and the `ServicePanelButton` component.

Reviewers:

- [x] @JanVoracek
- [x] @vasek17


vasek17:
I don't like the VersionPress title in `ServicePanel` component. I think it should stay in the parent component.

frontend/src/welcome-panel/WelcomePanel.tsx
```diff
@@ -0,0 +1,29 @@
...
+  onHide(e: React.MouseEvent): void;
+}
+
+const WelcomePanel: React.StatelessComponent<WelcomePanelProps> = ({ onHide }) => (
+  <div className='WelcomePanel welcome-panel'>
+    <a
+      className='welcome-panel-close'
+      href='#'
+      onClick={onHide}
+    >
```
vasek17: In this case (~60 chars) I think the attributes on one line are more readable.

frontend/src/service-panel/ServicePanel.tsx
```diff
@@ -0,0 +1,21 @@
...
+interface ServicePanelProps {
+  isVisible: boolean;
+  onButtonClick(e: React.MouseEvent): void;
+}
+
+const ServicePanel: React.StatelessComponent<ServicePanelProps> = ({ isVisible, onButtonClick }) => (
+  <div>
+    <Button onClick={onButtonClick} />
+    <VpTitle />
+    <VpInfo isVisible={isVisible} />
```
vasek17: `VpInfo` is not quite exact title for the component. Something like `Panel` will be better IMO.

frontend/src/service-panel/VpInfo.tsx
```diff
@@ -0,0 +1,49 @@
+import * as React from 'react';
+
+import * as classNames from 'classnames';
```
vasek17: One line up :)

frontend/src/service-panel/VpInfo.tsx
```diff
@@ -0,0 +1,49 @@
...
+}
+
+const VpInfo: React.StatelessComponent<VpInfoProps> = ({ isVisible }) => {
+  const systemInfoUrl = config.api.adminUrl + '/admin.php?page=versionpress/admin/system-info.php';
+  const wrapperClassName = classNames({
+    'ServicePanel-wrapper': true,
+    'ServicePanel-wrapper--hide': !isVisible,
+  });
+
+  return (
```
vasek17: This could be also extracted to 2 components. Maybe something like `panel/warning`, `panel/support`?

frontend/src/service-panel/ServicePanel.tsx
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react';
+
+import Button from './Button';
+import VpTitle from './VpTitle';
+import VpInfo from './VpInfo';
+import './ServicePanel.less';
```
vasek17: One empty line before the style import.

frontend/src/welcome-panel/WelcomePanel.tsx
Ishirak: Same as [here](https://github.com/versionpress/versionpress/pull/1105#discussion_r74811561).

Ishirak:
I don't think that having VersionPress title in the `ServicePanel` component causes any issues. In the current HTML structure it is kind of logical to have it there. A possible compromise could be something like this:

```jsx
<ServicePanel
  isVisible={this.state.displayServicePanel}
  onButtonClick={this.onServicePanelClick}
>
  <VpTitle />
</ServicePanel>
```

frontend/src/welcome-panel/WelcomePanel.tsx
borekb: @Ishirak It's not the same because that other example could not reasonably fit in a single line while this one could. But again, it's a small thing, either style is OK and will not make a huge difference to readability.

vasek17:
@Ishirak When I would want to change the main heading on the page I really would not look for it in the `ServicePanel` component.
:+1: for the proposed style.

PR #1112, opened 2016-08-26
Refactor of CommitsTable component

### LINKED ISSUE #1087: Use simple and clear React.js components

Ishirak:
Resolves (partialy) #1087.

Last major refactor of front end codebase.

Refactor of `HomePage` component will be the last piece to close #1087, but its refactor can only begin after this and #1110 are reviewed and merged.

Reviewers:

- [x] @JanVoracek 
- [x] @vasek17


JanVoracek:
I think that having a component for each scope (Post, Comment, WordPressUpdate, ...) is not very convenient as the scopes are dynamically defined by `schema.yml`. We can keep them for now but they will be removed after #1036 anyway.

vasek17:
@JanVoracek I am aware of that but for the scope of this PR I only refactored `getLinesFor...` methods into separate components without other changes to the functionality.

PR #1115, opened 2016-09-09
Refactor of HomePage component and directory structure

### LINKED ISSUE #1087: Use simple and clear React.js components

Ishirak:
Resolves #1087.

Last piece to close #1087 – major refactor of front end codebase will be over.

Reviewers:

- [x] @JanVoracek 
- [x] @vasek17


frontend/src/components/home/HomePage.tsx
```diff
@@ -136,7 +136,6 @@ export default class HomePage extends React.Component<HomePageProps, HomePageSta
             isLoading: false,
             displayUpdateNotice: false,
           });
-          this.checkUpdate();
```
vasek17: @Ishirak: Why is this line removed? If the checkUpdate is not called after fetching commits there could be short (<10s) inconsistency between the commits table and the CommitPanel.

Ishirak: You are right, I thought that in this case `setState()` would behave as an asynchronous method, as usual, but it doesn't, as I just checked. I will revert this change in the next commit.

PR #1116, opened 2016-09-09
Custom ini serializer

### LINKED ISSUE #912: Custom INI parser
`parse_ini_string()` is not entirely reliable and we already have quite a few workaround in our [IniSerializer](https://github.com/versionpress/versionpress/blob/3b161aa31d1e9639e60d6de45cc76f131259b964/plugins/versionpress/src/Storages/Serialization/IniSerializer.php) class. We're still encountering some new edge cases, e.g., #910, and it might be a good idea to create our own parser to avoid all the mess.

Maxxa:
Resolves #912 .

Reviewers:
- [ ] @JanVoracek

### DIFF
--- a/plugins/versionpress/src/Storages/Serialization/IniSerializer.php
+++ b/plugins/versionpress/src/Storages/Serialization/IniSerializer.php
@@ -158,16 +158,71 @@ class IniSerializer
      */
     public static function deserialize($string)
     {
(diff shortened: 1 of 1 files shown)


borekb:
We'll take a fresh look in the 5.0 timeframe, thanks for the first draft of a custom implementation.

PR #1134, opened 2016-10-10
Fix jQuery require for WebUI popover

mccahan:
Resolves # .

This resolves a JS console error I was encountering on a theme that deregisters jQuery and re-registers it to run in the footer to defer jQuery loading.

Reviewers:

- [ ] @versionpress/core-devs
Resolves issues with themes that use their own jQuery instead of WordPress' built-in version.


borekb:
@mccahan Hi, sorry for a late reply, was the proposed code change incorrect or is there another reason for closing the PR? Thanks.

mccahan:
@borekb Was fixed in a different pull request (see: https://github.com/versionpress/versionpress/commit/cae24e61d9892492edc80f834d47243b2149c2d8 ) - I think I'd re-submitted using `array()` notation instead of `[]` since I wasn't 100% positive what your PHP version requirements were at the time.

borekb:
Oh great, thanks for the info!

PR #1137, opened 2016-10-10
Autocomplete: Capital P, dangit!

### LINKED ISSUE #1136: Autocomplete: Capital P, dangit!

vasek17:
Resolves #1136.

Reviewers:

- [x] @JanVoracek 
- [x] @borekb

### DIFF
--- a/plugins/versionpress/src/Utils/AutocompleteUtils.php
+++ b/plugins/versionpress/src/Utils/AutocompleteUtils.php
@@ -104,15 +104,18 @@ class AutocompleteUtils
     }
 
     /**
+     * Capitalize the first word of a string.
+     * Also fix the capitalization of WordPress and VersionPress.
+     *
      * @param $string
      * @return string
      */
     private function capitalize($string)
     {
         $capitalizedString = Strings::firstUpper($string);
-        return str_replace(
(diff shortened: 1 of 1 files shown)


plugins/versionpress/src/Utils/AutocompleteUtils.php
```diff
@@ -102,4 +102,18 @@ private static function getDateConfig()
             'type' => 'date'
         ];
     }
+
+    /**
+     * @param $string
+     * @return string
+     */
+    private function capitalize($string)
```
borekb: This function could use a description (it's not that obvious what it's doing).

plugins/versionpress/src/Utils/AutocompleteUtils.php
```diff
@@ -102,4 +102,18 @@ private static function getDateConfig()
...
+    {
+        $capitalizedString = Strings::firstUpper($string);
+        return str_replace(
+            ['Wordpress', 'wordpress', 'Versionpress', 'versionpress'],
```
borekb: The impl seems too specific. Wouldn't it be better to match "wordpress" and "versionpress" case-insensitively and replace with proper forms?

PR #1143, opened 2016-10-13
Fix zip compression

### LINKED ISSUE #1138: Unable to unzip 4.0-alpha1
I just downloaded versionpress-4.0-alpha1 and am getting an error when unzipping.  Have tried it several different ways.

screenshot 2016-10-10 15 03 13

Ishirak:
Resolves #1138.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -375,6 +375,7 @@ function isRelative(sitePath) {
 
 /**
  * Compress the build into a zip file.
+ * We don't use gulp-zip because of this bug: https://github.com/sindresorhus/gulp-zip/issues/79
  * Inspiration from https://github.com/sindresorhus/gulp-tar/blob/cbe4e1df44fdc477a3a9743cfb62ccb999748c16/index.js
  */
 function compress(filename) {
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -7,7 +7,6 @@ var del = require('del');
 var shell = require('gulp-shell');
 var shelljs = require('shelljs/global');
 var rename = require('gulp-rename');
-var zip = require('gulp-zip');
 var replace = require('gulp-replace');
 var fs = require('fs');
 var path = require('path');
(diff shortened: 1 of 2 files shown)


gulpfile.js
```diff
@@ -370,4 +371,52 @@ gulp.task('idea', 'Setup project files for IDEA / PhpStorm', function () {
 
 function isRelative(sitePath) {
     return path.normalize(sitePath) != path.resolve(sitePath);
+}
+
+/**
+ * Compress the build into a zip file.
```
borekb: There should be an explanation here why we're not using the `gulp-zip` package – a link to some issue or something.

borekb:
@Ishirak, can you link to some `gulp-zip` issue or something similar which would explain what was the problem? Did you encounter anything useful during your research?

Ishirak:
@borekb in the process of decompressing the zip file on the Mac I've encountered following error couple of times:

> symlink error: File name too long

In the process of fixing this error I've tried to update all of the packages in the `package.json`. No succes.

Next I've tried to compress the build using default Mac tools. Succes – problem had to be in the `gulp-zip` package.

So I've tried to use some alternatives to the `gulp-zip`, e.g. `gulp-gzip`, `gulp-tar` or `archiver`. In every single one I've tried I've encountered another issue, e.g. no support for single directory compression, compression of single files etc.

Only [gulp-tar](https://github.com/sindresorhus/gulp-tar) was almost perfect – written by the same author as [gulp-zip](https://github.com/sindresorhus/gulp-zip), but used different compression algorithm. Its only downside was the `.tar` suffix. So I've used `gulp-tar` as the source for the `compress()` function and changed its behaviour to support our use case – `.zip` suffix.

borekb:
Wow, just to generate a stupid ZIP. 👍

vasek17:
It is a bug in `gulp-zip` package, I reproduced and extracted the minimal example.
Related issue: https://github.com/sindresorhus/gulp-zip/issues/79.

PR #1178, opened 2016-12-30
1175 processinsertquery

### LINKED ISSUE #1175: "processInsertQuery" relies on "wpdb->insert_id"
The `WpdbMirrorBridge->processInsertQuery` method relies on `wpdb->insert_id` as the id for inserting into the `vp_id` table. With plugin integration, not all tables will use an auto increment value that `wpdb` will pickup. I ran into this when integrating with Gravity Forms. Their `form_meta` table just uses the `id` from the `form` table. Maybe as a fall back, lookup the identified `id` in the schema.yml file and use that if `wpdb` fails to give an id. Or maybe rely more on the schema identified id during inserts.

Here is what I hacked together just so I can continue work on integrating Gravity Forms. Sure something more robust will be needed. 

```php
if ($this->database->insert_id == 0) {
    $idColumnNames = $parsedQueryData->idColumnsNames[0];
    $id = $parsedQueryData->data[0][$idColumnNames];
} else {
    $id = $this->database->insert_id;
}
```

unclhos:
Resolves #1175 

When table does not use auto-increment value for an id WordPress returns a "0". When this happens, use the field defined in the schema.yml file as the id.


unclhos:
Sorry, some old commits in this branch. I branched from my master and not yours.

PR #1179, opened 2016-12-30
1176 plugindiscovery

### LINKED ISSUE #1176: Plugin definition discovery: local location outside of <plugin>/.versionpress
https://github.com/versionpress/versionpress/issues/1036 implements a discovery method that expects plugin definition files (`schema.yml`, `actions.yml` etc.) to be located at `<some-plugin>/.versionpress`. There should also be an external location from where the definitions could be loaded, two initial candidates being:

```
wp-content/plugins/.versionpress
wp-content/vpdb/.versionpress (or some other folder name)
```

I like the `plugins` folder better because not everything in a plugin definition is tied to a DB and `vpdb` is very database-specific.

Inside it, it would probably just be directories for plugins so the complete example would look like:

```
wp-content/plugins/.versionpress
    akismet/
        schema.yml
        actions.yml
    woocommerce/
        schema.yml
        actions.yml
    ...
```

After this is implemented, there will be an online repo of these definitions, sort of like DefinitelyTyped for TypeScript. VersionPress will be downloading plugin definitions from that repo to this local structure. See https://github.com/versionpress/versionpress/issues/1243.

---

🏁 PR https://github.com/versionpress/versionpress/pull/1246

unclhos:
Resolves #1176 

Adds a new location for plugin discovery when versionpress definitions are not included with plugin.


unclhos:
Sorry, some old commits in this branch. I branched from my master and not yours.

PR #1187, opened 2017-01-24
Added VP_WP_CLI_BINARY option to vp config

### LINKED ISSUE #1186: Adding VP_WP_CLI_BINARY option to vp config
I am running  VersionPress on my shared cPanel hosting. The problem is that I was not able to create global `wp` alias for WP-CLI. I had to modify VersionPress files to include full path to WP-CLI binary. This changes will be lost with next update. 

The solution would be to create a new `vp config` option for the WP-CLI path (simmilar to `VP_GIT_BINARY`) so you can change the WP-CLI path just by running something like
 `wp vp config VP_WP_CLI_BINARY '/custom/path/to/wp-cli.phar'`

LorincJuraj:
Resolves #1186 

Creates new [config constant](https://docs.versionpress.net/en/getting-started/configuration#versionpress-config-constants) which gives an option to set custom path to [WP-CLI](http://wp-cli.org/). 
Works same way as _VP_GIT_BINARY_:

`wp vp config VP_WP_CLI_BINARY '/custom/path/to/wp'`

### DIFF
--- a/plugins/versionpress/bootstrap.php
+++ b/plugins/versionpress/bootstrap.php
@@ -38,6 +38,20 @@ if (!defined('VP_GIT_BINARY')) {
     define('VP_GIT_BINARY', 'git');
 }
 
+if (!defined('VP_GIT_BINARY')) {
+    /**
+     * Absolute path to the git executable. Useful if it's not in PATH.
+     */
+    define('VP_GIT_BINARY', 'git');
+}
+
+if (!defined('VP_WP_CLI_BINARY')) {
+    /**
+     * Absolute path to the WP-CLI executable. Useful if it's not in PATH.
+     */
+    define('VP_WP_CLI_BINARY', 'wp');
+}
+
 if (!defined('VERSIONPRESS_GUI')) {
     /**
      * Which GUI to use. Used mainly for testing.
--- a/plugins/versionpress/bootstrap.php
+++ b/plugins/versionpress/bootstrap.php
@@ -38,13 +38,6 @@ if (!defined('VP_GIT_BINARY')) {
     define('VP_GIT_BINARY', 'git');
 }
 
-if (!defined('VP_GIT_BINARY')) {
-    /**
-     * Absolute path to the git executable. Useful if it's not in PATH.
-     */
-    define('VP_GIT_BINARY', 'git');
-}
-
 if (!defined('VP_WP_CLI_BINARY')) {
     /**
      * Absolute path to the WP-CLI executable. Useful if it's not in PATH.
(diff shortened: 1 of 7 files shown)


borekb:
Thanks, @LorincJuraj, we'll look at it soon.

plugins/versionpress/bootstrap.php
```diff
@@ -38,6 +38,20 @@
     define('VP_GIT_BINARY', 'git');
 }
 
+if (!defined('VP_GIT_BINARY')) {
```
borekb: Duplicates the block above, probably an oversight.

plugins/versionpress/src/Cli/vp.php
```diff
@@ -44,11 +44,12 @@ class VPCommand extends WP_CLI_Command
      *
      * <constant>
      * : The name of the constant to set.
-     *   VP_GIT_BINARY:   Absolute path to the git binary.
-     *   VP_PROJECT_ROOT: Absolute path to the root of your project (typically
-     *                    where is the .git directory)
-     *   VP_VPDB_DIR:     Absolute path to directory where VersionPress saves
-     *                    versioned database data.
+     *   VP_GIT_BINARY:     Absolute path to the git binary.
+     *   VP_WP_CLI_BINARY:  Absolute path to the WP-CLI binary.
```
borekb: The new implementation maybe even allows defining `VP_WP_CLI_BINARY` as `'php wp-cli.phar'`; I can imagine that could be useful sometimes. Is that correct?

plugins/versionpress/bootstrap.php
LorincJuraj: @borekb you are right. It was an oversight. Fixed now c9c777942a9e88efbc58ad60d8b4e717d329fdf1 . Thanks.

plugins/versionpress/src/Cli/vp.php
LorincJuraj: I have just now realized that you can not use it on command line, because the command line is using the wp-cli so if it is not working properly, the command `wp vp config VP_WP_CLI_BINARY '/custom/path/to/wp-cli'` will not work. But maybe there is a way to make `vp config` work without wp-cli so you could run just `vp config VP_WP_CLI_BINARY '/custom/path/to/wp-cli'`.

But the `VP_WP_CLI_BINARY` is still useful because you can set it by manually editing wp-config.php.

JanVoracek: It does not allow `php wp-cli.phar` as it can be only an absolute path. `wp vp config` checks whether the file exists.

PR #1208, opened 2017-05-10
Switch to LF line endings for INI files

### LINKED ISSUE #728: Switch to LF line endings for INI files
Our INI files are [strictly CRLF](https://github.com/versionpress/versionpress/blob/9d5a7173a7dd36f104883a7d4ef823fe4e8c36f6/plugins/versionpress/src/Utils/IniSerializer.php#L10) today and I can't remember why - it was certainly part of the original INI "standard" because this format was developed on Windows but I'm not sure if there were other reasons why we decided to use CRLF. For example, PHP function [`parse_ini_string()`](http://php.net/manual/en/function.parse-ini-string.php) that we use internally seems to support LF line endings just fine, and we generally much-prefer LF even on Windows (the project is strictly LF, see #177).

Apart from the fact that even Windows users on our team don't like CRLF line endings, it's constantly causing subtle issues like tests failing because the strings contain different line endings or something stupid like that. Also, when Git inserts conflict markers in a file, it uses LF which is weird when inserted into a CRLF INI file. Again, it feels safer and more "right" to have everything in LF.

JanVoracek:
Resolves #728

### DIFF
--- a/plugins/versionpress/src/Storages/Serialization/IniSerializer.php
+++ b/plugins/versionpress/src/Storages/Serialization/IniSerializer.php
@@ -276,7 +276,7 @@ class IniSerializer
 
     private static function outputToString($output)
     {
-        return implode("\r\n", $output);
+        return implode("\n", $output);
     }
 
     private static function preserveNumbers($iniString)
--- a/plugins/versionpress/src/Utils/StringUtils.php
+++ b/plugins/versionpress/src/Utils/StringUtils.php
@@ -26,6 +26,17 @@ class StringUtils
     }
 
     /**
+     * Ensures LF line endings in a string
+     *
+     * @param string $str LF or CRLF line endings
+     * @return string LF line endings
+     */
+    public static function ensureLf($str)
+    {
+        return str_replace("\\r\\n", "\\n", $str);
+    }
+
+    /**
      * Converts given noun to plural form.
      * Note: It's very, very simplified!
      *
--- a/plugins/versionpress/src/Utils/StringUtils.php
+++ b/plugins/versionpress/src/Utils/StringUtils.php
@@ -26,17 +26,6 @@ class StringUtils
     }
 
     /**
-     * Converts LF string to CRLF string
-     *
-     * @param string $str LF line endings
-     * @return string CRLF line endings
-     */
-    public static function crlfize($str)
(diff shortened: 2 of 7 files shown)


plugins/versionpress/src/Storages/Serialization/IniSerializer.php
```diff
@@ -276,7 +276,7 @@ private static function getReplacedEolString($str, $direction)
 
     private static function outputToString($output)
     {
-        return implode("\r\n", $output);
+        return implode("\n", $output);
```
JanVoracek: `parse_ini_string` supports both LF and CRLF

borekb: It does. I've added a test to make sure.

borekb:
I've pushed the discussed updates, could you please review it @JanVoracek?

plugins/versionpress/src/Utils/StringUtils.php
```diff
@@ -26,14 +26,14 @@ public static function verbToPastTense($verb)
...
-     * @param string $str LF line endings
-     * @return string CRLF line endings
+     * @param string $str LF or CRLF line endings
+     * @return string LF line endings
      */
-    public static function crlfize($str)
+    public static function ensureLf($str)
     {
-        return str_replace("\n", "\r\n", str_replace("\r\n", "\n", $str));
+        return str_replace("\\r\\n", "\\n", $str);
```
JanVoracek: > Ensures LF line endings in a string.

This is not true. Because of escaping it replaces string '\r\n' (`ASCII 0x005C 0x0072 0x005C 0x006E`) with '\n' (`ASCII 0x005C 0x006E`). Is it really what we want? I thought we want to replace CRLF (`ASCII 0x000D 0x000A`) with LF (`ASCII 0x000A`).

borekb: But it's good that I covered it with unit tests.. OMG 🤦‍♂️.

borekb: Fixed in https://github.com/versionpress/versionpress/pull/1208/commits/9ac16030da390ad39296c83512f24f92d13447e8.

Issue #1213, opened 2017-05-12
Update dev setup and dependencies for 4.0-beta


borekb:
I'd like to review our dev setup for 4.0-beta. Some of the things already have their own issues but it might be easier to do all of the updates in a fewer number of PRs (probably still multiple).

- [x] Update VP project (PHP). Done mostly in Dockerized dev setup, https://github.com/versionpress/versionpress/pull/1227.
- [x] Update frontend project (TypeScript / React): PR https://github.com/versionpress/versionpress/pull/1214
- [x] ~~npm -> Yarn~~ won't happen, see https://github.com/versionpress/versionpress/issues/1142#issuecomment-309294152
    - [x] Revert parts of https://github.com/versionpress/versionpress/pull/1214 in a separate PR; frontend already switched to Yarn. _done in Dockerized dev setup #1227_
- [x] Update .idea template files #1144
    - [x] This has also happened in Dockerized dev setup #1227
- [x] PhpStorm inspections _done in Dockerized dev setup #1227_
- ~~Adopt PHPStan~~ see separate issue #1210
- ~~Review and update Dev-Setup.md~~ -> part of Dockerized dev setup #1227
- ~~Gulp -> npm scripts?~~ Won't happen at this point.

Issue #1142, opened 2016-10-12
npm -> Yarn


borekb:
I've quickly tried [Yarn](https://yarnpkg.com/) and it was pretty impressive: our current dependencies were installed in about 10 seconds, compared to over 2 minutes with "good old npm". I see no reason not to switch (Yarn is [pretty compatible with npm](https://yarnpkg.com/en/docs/migrating-from-npm)).

Ishirak:
I'm all in to use Yarn, but in #1140 @vasek17 introduced `types` instead of `typings` and according to [this issue](https://github.com/yarnpkg/yarn/issues/825) Yarn doesn't yet support this type of definition files. I'm sure that Yarn will suport it eventually, so I'd wait until it does so.

vasek17:
What exactly do you mean by switch "npm -> yarn"? Adding a `yarn.lock` file to the repository? Changing [these](https://github.com/versionpress/versionpress/blob/203257484e5855e34cbb78a3e929019ff86e94f5/gulpfile.js#L273) [lines](https://github.com/versionpress/versionpress/blob/203257484e5855e34cbb78a3e929019ff86e94f5/gulpfile.js#L290) (Running `npm install` from gulp scripts)?

Yarn is compatible with npm (well, will be in future versions) so I would leave it on user to choose `npm install` or `yarn install` (and not enforce using `yarn`).

borekb:
I'd switch to Yarn completely as I don't see any benefit of offering a choice (and maintaining two sets of lock files, documentation etc.).

There are some blockers, though, most prominently:

- https://github.com/yarnpkg/yarn/issues/825 as mentioned by @Ishirak 
- https://github.com/yarnpkg/yarn/issues/853 (preinstall / postinstall scripts)

vasek17:
:+1:

vasek17:
I still like more the optional way.  Anyone will be able to run `yarn install` instead of `npm install` but if he don't like/don't have `yarn` installed, it would still work wild "good old npm".

borekb:
BTW, https://github.com/yarnpkg/yarn/issues/825 is fixed in Yarn 0.18, https://github.com/yarnpkg/yarn/issues/853 is now addressed as well.

borekb:
[...]

PR #1227, opened 2017-05-21
1041 Dockerized dev setup

### LINKED ISSUE #974: Update project dependencies
According to `composer outdated` (below) we use several outdated libraries. Some of them are crucial for VersionPress (e.g. `symfony/process`) or heavily used (e.g. `nette/utils`). Also, there are some dependencies we no longer use, like `comment-stripper`. We should review these, too. 

screen shot 2016-04-24 at 19 53 25

### LINKED ISSUE #1041: Dockerized development environment
Our [Dev-Setup](https://github.com/versionpress/versionpress/blob/master/docs/Dev-Setup.md) is quite complex and it would be better to provide pre-configured environment, built on Docker (which we love!). @JanVoracek already tried to prototype it and got quite far, however, Xdebug was not working properly. We should look at it again.

---

🏁 PR #1227

### LINKED ISSUE #1144: Update .idea template files
It seems that PhpStorm updated its format a bit (I'm currently running 2016.3), plus we might have some older dependencies in the IML file:

image

Let's review the `.idea` template files.

borekb:
Resolves #1041 (Dockerized development environment)
Resolves #1144 (Update .idea template files)
Resolves #974 (Update project dependencies)
Resolves parts of #1213 (Update dev setup and dependencies for 4.0-beta)

This is not perfect yet but already implements a lot of work so let's merge it.

## Summary

- Local development environment is powered by Docker, no need to set up local web server, database server or anything.
- Tests run on Docker as well. Selenium tests are now as easy to run as unit tests.
- Some VersionPress code had to be updated to work in a read-only environment (`wp-content/plugins/versionpress` is a read-only mount), for example, there's a new `wp-content/vpcache` folder instead of `wp-content/plugins/versionpress/temp`.
- `Dev-Setup.md` rewritten and simplified, `Testing.md` incorporated into it.
- Dependencies update and general project cleanup:
    - PhpStorm instructions and project files updated to 2017.2.
    - `.editorconfig` files unified.
    - PHP dependencies updated.

## Known issues

- Most test suites pass when run individually but fail when run together (`npm run tests`). This needs further inspection.
- Some end2end tests with Selenium workers fail.

## Future work

- Parametrize runs with WP-CLI or Selenium workers, or even do both at the same time (would require bigger change).
    - Technically, TestConfig / test-config.yml and TestRunnerOptions could be unified. Any parameter from test-config should be switchable via CLI argument.
- Travis CI.
- Selenium tests can be observed using VNC viewer, not that useful but it's fun to watch.

---

TODO list:

- [x] Build a basic Docker-based environment:
    - [x] WordPress latest on PHP 5.6
    - [x] WP-CLI and Git inside it
    - [x] MySQL
    - [x] Adminer
- [x] Map `<repo>/plugins/versionpress` into the container
- [x] PhpStorm setup, document debugging
- [x] Git on the host should be able to inspect site repo
- [x] DB tools on the host should be able to connect to the MySQL database
- [x] Turn off `xdebug.remote_autostart` for performance reasons
- [x] Pre-install WordPress via WP-CLI (currently, web installation needs to be finished)
- [x] Create npm scripts for tasks like starting up the stack or starting fresh
- [x] Update Dev-Setup.md
    - [x] Build section
- [x] Run tests in a container
    - [x] Unit tests
    - [x] Full integration tests (Selenium, maybe switch to Chrome headless)
        - [x] Basic docker-compose-based testing stack. _Done in https://github.com/versionpress/versionpress/pull/1227/commits/6843b874b21fef3ea72f5e1958d40c871ecba213._
        - [x] Incorporate all the other tests (currently, only `PublicWebTest` is run). There will very likely be many smaller issues, for example, `WpConfigSplitterTest` tests are failing because file system is now read-only.
        - [x] `WpAutomation` currently does tricky handling of users and permissions (root vs. www-data), this could most likely be simplified.
        - ~~Make both Selenium and WP-CLI workers work easily at the same time (right now, it is a test-config.yml value that needs to be switched manually).~~ Will be a separate PR.
        - [x] Selenium host should be configurable via `test-config.yml`.
        - [x] Testing section of `Dev-Setup.md` will need some cleanup.
        - [x] In the end, don't forget to get rid of the temporary `selenium-tests`.
    - [x] Integrate `Testing.md` into `Dev-Setup.md`
- [x] Sort out `frontend` dev
- [x] Smaller things while we're at it:
    - [x] Unify .editorconfig files
    - [x] Consider `docker-compose up -d`


dev-env/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+  echo 'sudo -u www-data /bin/wp-cli.phar "$@"'; \
+} > /bin/wp
+RUN chmod +x /bin/wp-cli.phar && chmod +x /bin/wp
+
+# Xdebug
+# Adapted from https://github.com/johnrom/docker-wordpress-wp-cli-xdebug/blob/8b87351f9b65b95734fd726e97deff45ec8c8dfc/Dockerfile
+RUN yes | pecl install xdebug \
+    && echo "zend_extension=$(find /usr/local/lib/php/extensions/ -name xdebug.so)" > /usr/local/etc/php/conf.d/xdebug.ini \
+    && echo "xdebug.remote_enable=1" >> /usr/local/etc/php/conf.d/xdebug.ini \
+    && echo "xdebug.remote_autostart=1" >> /usr/local/etc/php/conf.d/xdebug.ini \
```
JanVoracek: This is what makes the responses slow. See http://stackoverflow.com/a/19273025/1243495

borekb:
The PR currently recommends Yarn but we'll stay with npm, see Yarn was initially used in this PR, move back to npm, see https://github.com/versionpress/versionpress/issues/1142. I've added todo item to switch from Yarn to npm in this PR.

borekb:
Major progress on dockerizing integration tests in https://github.com/versionpress/versionpress/pull/1227/commits/6843b874b21fef3ea72f5e1958d40c871ecba213. I'm very happy that running full-blown Selenium tests is now as easy as unit tests. Compare:

- `docker-compose run unit-tests`
- `docker-compose run selenium-tests`

It won't get much better than that :)

Next steps:

- [ ] Incorporate all the other tests (currently, only `PublicWebTest` is run).
- [ ] `WpConfigSplitterTest` "unit" tests are failing because they write to file system which is now read-only.
- [ ] `WpAutomation` currently does tricky handling of users and permissions (root vs. www-data), this could most likely be simplified.
- [ ] Choose a better name for the test site (currently, the legacy name `vp01` used).
- [ ] Testing section of `Dev-Setup.md` will need some cleanup, documenting VNC viewer setup would be cool.

borekb:
@JanVoracek I'm seeking feedback on certain things from https://github.com/versionpress/versionpress/pull/1227/commits/6843b874b21fef3ea72f5e1958d40c871ecba213 please:

1. Does it generally look right? For example, one slightly controversial thing is mounting the project folder to `/var/www/html/vp01/wp-content/plugins/versionpress` inside the containers: it makes it very convenient to edit files on the host OS and immediately see the changes inside containers, and it's also safe as the the mount is read-only but it also means that Docker will lock those files to `root` (impossible to change to www-data) and also, technically speaking, the contents of `/var/www/html/vp01/wp-content/plugins/versionpress` is slightly different than what `gulp build` exports and users install. I think those are reasonable trade-offs but would like to know your opinion.
2. Is this OK?
    > `VERSIONPRESS_TEMP_DIR` changed from `VERSIONPRESS_PLUGIN_DIR/temp` to `WP_CONTENT_DIR/vpcache` as the plugin directory is now mounted as a read-only file system.
    
    I've chosen the name `vpcache` as it seems that the main use of this folder in v4 is caching definition files (not only that but I think `vpcache` is still quite suitable). So in `WP_CONTENT_DIR`, VersionPress currently creates two folders: `vpdb` and `vpcache`.
3. If you try to follow the instructions, do they work?

JanVoracek:
Progress on tests:

- [ ] End2End
- [x] GitRepositoryTests
- [ ] LoadTests
- [ ] Selenium
- [x] SqlTests
- [x] StorageTests
- [x] SynchronizerTests
- [x] Unit
- [ ] Workflow

borekb:
I've updated the original description to provide all the details on this PR. It is large enough so let's merge it!

image

Issue #1243, opened 2017-06-20
Online repository for plugin definitions


borekb:
After local discovery of plugin definitions in `WP_PLUGIN_DIR/<plugin>/.versionpress` since https://github.com/versionpress/versionpress/issues/1036 and `WP_CONTENT_DIR/.versionpress/<plugin>` since https://github.com/versionpress/versionpress/issues/1176, the final step is to have an online repository of the definitions. It will be open to anyone to contribute (not just plugin authors) and used by VersionPress transparently.

It would work like this:

First, a user goes to the _Add Plugins_ screen and sees whether the plugin has definitions or not:

![image](https://user-images.githubusercontent.com/101152/27338124-a1b09186-55d4-11e7-91d1-31276b339774.png)

If the plugin is compatible, the user can install it. VersionPress will download the definition to `WP_CONTENT_DIR/.versionpress/<plugin>` (or whatever https://github.com/versionpress/versionpress/issues/1176 implements) and use it.

When the plugin is uninstalled, it is no different from how local-only definitions behave.

The online repository of definitions will probably take a form of:

- **GitHub repo** for community contribution, sort of like [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped). Probably `versionpress/definitions`.
    - DefinitelyTyped is hitting a bit of a scaling issue, e.g., cloning the repo is slow. An alternative would be to have a GitHub organization for definitions and then repositories in it like `versionpress-definitions/akismet`, `versionpress-definitions/hello-dolly` etc. But we don't need to complicate it right now.
- Small **web service hosted by us** e.g. to avoid GitHub API rate limits (60 requests per hour are allowed for unauthenticated requests and that can be consumed easily by sites with multiple plugins). This web service could also provide things like definition search etc. but the prime reason is technical.

Some things that will need thinking through are:

- Repository structure.
- How to deal with plugin versions (we probably won't in v1 of this).

tannerchung:
This would be awesome!  Keen on seeing the yml for theme support!

PR #1246, opened 2017-06-23
Plugin definition discovery

### LINKED ISSUE #1176: Plugin definition discovery: local location outside of <plugin>/.versionpress

JanVoracek:
Resolves #1176 

VP now prefers loading definition files from `wp-content/.versionpress`. I think it should be enough for the first iteration.


plugins/versionpress/src/Actions/ActionsDefinitionRepository.php
```diff
@@ -31,14 +31,19 @@ public function getAllDefinitionFiles()
     public function saveDefinitionForPlugin($pluginFile)
     {
         $pluginSlug = basename(dirname($pluginFile));
-        $actionsFile = WP_PLUGIN_DIR . '/' . $pluginSlug . '/.versionpress/actions.yml';
-
-        if (!is_file($actionsFile)) {
+        $defaultActionsFile = WP_CONTENT_DIR . '/.versionpress/' . $pluginSlug . '/actions.yml';
+        $pluginDirActionsFile = WP_PLUGIN_DIR . '/' . $pluginSlug . '/.versionpress/actions.yml';
+
+        if (is_file($defaultActionsFile)) {
```
borekb: Maybe the IF construction which is mostly repeated in all three files could be refactored to something like `$actionsFile = PluginDefinitionDiscovery::getPath($pluginSlug, 'actions.yml')`. It would also allow to avoid words like "default" and "effective" which I'm not sure are entirely fitting.

borekb:
Nice refactoring 👍 

Before we merge, we should discuss the final structure of paths:

1. Do we like the `WP_CONTENT_DIR/.versionpress/<plugin-slug>` path? Some alternatives:
    1. `WP_PLUGIN_DIR/.versionpress/<plugin-slug>` – would support `WP_THEME_DIR/.versionpress/<theme-slug>` in a natural way. (Please also see note on themes below.)
    2. `WP_CONTENT_DIR/versionpress-definitions/<plugin-slug>` (or something shorter than `versionpress-definitions` like `vpdefs` or `@vp`). The point here is to avoid the "dot name" which could maybe cause some issues in some cases (for example, `ls` does not list it by default, some backup plugin might iterate directories and skip the dot files, etc.).
    3. `WP_CONTENT_DIR/.versionpress/plugins/<plugin-slug>` to make room for themes in the future. Two notes here:
        - Maybe it's out of scope to think about themes now. We can delay this decision.
        - On a typical site, plugins dominate (many plugins per one active theme). Maybe it would be OK to have plugins at the top-level directory (`WP_CONTENT_DIR/.versionpress`) and put themes to a specially named one like `WP_CONTENT_DIR/.versionpress/@themes`. But it's just a thought now, I'm not sure I like it better.
1. What about core definitions? Do we like the path `WP_PLUGIN_DIR/versionpress/.versionpress`? I personally think it would be more consistent to put it in a subfolder like `.versionpress/wp-core` (of course making it unique so that it doesn't conflict with any plugin slug).
2. On a related note, do we want to support `WP_PLUGIN_DIR/versionpress/.versionpress` as one of the discovery locations? For example, if we wanted to ship WooCommerce support out of the box, it would give us the opportunity to use the `WP_PLUGIN_DIR/versionpress/.versionpress/woocommerce` folder.
    - An alternative would be to use these folders in our source code but do not use it at runtime as a dicovery location. VersionPress would need to copy the files from there to `WP_CONTENT_DIR/.versionpress/<plugin-slug>` on activation. Again, just an idea, I'm not sure I like it that much.

JanVoracek:
1.i.: It looks ok but what about WP core?
1.ii.: I think `.versionpress` shouldn't be a problem.  It's similar to `.git` – it will break the site if someone performs backup & restore without `.git`.
1.iii.: I think this will be the best. `WP_CONTENT_DIR/.versionpress/plugins/<plugin-slug>` and `WP_CONTENT_DIR/.versionpress/themes/<stylesheet>` look fine.
2.: Subfolders in `WP_PLUGIN_DIR/versionpress/.versionpress` 👍 Maybe same structure as 1.iii.?
3.: We can drop the support of `WP_PLUGIN_DIR/versionpress/.versionpress` at any time. Let's keep it for now.

borekb:
After some more internal discussion, the decision is:

- The path will be `WP_CONTENT_DIR/.versionpress/plugins/<plugin-slug>` for plugins and `WP_CONTENT_DIR/.versionpress/themes/<stylesheet>` for themes.
    - We'll go with the "dot name" `.versionpress`.
- WP core definitions will not be updatable via `WP_CONTENT_DIR` just yet. If we support that in the future, it will probably a sibling folder to `/plugins` and `/themes` like `/core` or `/wp-core`.
- `WP_PLUGIN_DIR/versionpress/.versionpress` will not be a third discovery location, it will only be used for WP core definitions (and even that might change in the future; WP core definitions could be downloaded from the [online repo](https://github.com/versionpress/versionpress/issues/1243)).

pavelevap:
I would suggest structure similar to WordPress `languages` folder (everyone is familiar with it).

1. i. `WP_PLUGIN_DIR` is directory for plugins, we probably should not use it for definitions, it could lead to misunderstanding. `WP_CONTENT_DIR` is better in this case.
ii. This path can leads to problems with future support for themes.
iii. I agree that `WP_CONTENT_DIR/.versionpress/plugins/<plugin-slug>` is the best way. I am not sure if dot can lead to any problems, maybe it can be used without dot `WP_CONTENT_DIR/versionpress/plugins/<plugin-slug>` or change the whole name (no personal preference).

2. `WP_PLUGIN_DIR/versionpress/.versionpress` is OK, I guess. Core files can be also mirrored directly in `WP_CONTENT_DIR/.versionpress/` (without special subfolder). But there are related issues, see point 3.

3. How will be definitions stored and updated? There will be some repo (server, something...) to store definitions and allow users to contribute. For example support for WooCommerce. When plugin authors create their own definitions, they can ship it inside their plugin package `WP_PLUGIN_DIR/woocommerce/.versionpress/`. Or they can submit definitions to planned repo and VersionPress will distribute it into `WP_PLUGIN_DIR/versionpress/.versionpress/plugins/woocommerce` or `WP_CONTENT_DIR/versionpress/plugins/woocommerce`? Using `WP_PLUGIN_DIR` is not very good because definitions should be able to update more frequently than core plugin files and for this case is probably better `WP_CONTENT_DIR`. But on the other hand, when we distribute it into `WP_CONTENT_DIR`, we should also give developers some other place or hook to allow them modify these files without danger of automatic updates (safe place which will be somehow discoverable by VersionPress, but never modified automatically). There is a special filter (often used) for example for `languages` to allow modification without rewriting original files, [load_textdomain_mofile](https://developer.wordpress.org/reference/hooks/load_textdomain_mofile/).

borekb:
@pavelevap The plan for 3 is to download to `WP_CONTENT_DIR`. This takes precedence over the default definitions shipped with a plugin.

Filters are a good idea but we should probably wait until there's a specific need for it.

PR #1251, opened 2017-06-29
'edit' -> 'update' for most actions

### LINKED ISSUE #1120: 'edit' -> 'update' action verb
All database scopes in `actions.yml` define at least the `create`, `edit` and `delete` actions. The `edit` verb was probably chosen because posts were the first scope we ever worked on, however, the `update` verb is better and we should switch to it before 4.0 is released and people start defining `edit` actions.

This will be a breaking change.

---

🏁 PR https://github.com/versionpress/versionpress/pull/1251

borekb:
Resolves #1120 

All legacy `edit` actions converted to `update`. There are some cases where both actions are still valid, e.g., plugin and themes have both the `update` (to a new version) and `edit` (plugin/theme files) actions.

The change is backwards compatible: if the `edit` action comes to `TrackedChangeInfo` and the scope doesn't have a proper description for it, it is upgraded to the `update` action. See https://github.com/versionpress/versionpress/pull/1251/commits/fb432e6ecc231ec59fe925b5535222cb89887b5c.


borekb:
Thanks for the update, @JanVoracek, I've updated the PR description to indicate that this change is backwards compatible.

PR #1255, opened 2017-07-03
--allow-root for nested WP-CLI commands

### LINKED ISSUE #1049: Apply-changes doesn't pass allow-root
WP-CLI complains when running as root 

>  Error: YIKES! It looks like you're running this as root. You probably meant to run this as the user that your WordPress install exists under.
> 
> If you REALLY mean to run this as root, we won't stop you, but just bear in mind that any code on this site will then have full control of your server, making it quite DANGEROUS.
> 
> If you'd like to continue as root, please run this again, adding this flag:  --allow-root
> 
> If you'd like to run it as the user that this site is under, you can run the following to become the respective user:
> 
> ```
> sudo -u USER -i -- wp <command>
> ```

Their "sudo" work around seems to be buggy too. 
"wp vp apply-changes" doesn't pass the "--allow-root" flag.

> root@host [2570 15:10:31 /******_/**_*****]# wp vp apply-changes --allow-root
> Error: Maintenance mode couldn't be switched. Details:
> 
> Error: YIKES! It looks like you're running this as root. You probably meant to run this as the user that your WordPress install exists under.

JanVoracek:
Resolves #1049.

I think we can add the switch to all commands, no matter they were run under root or not. Commands running under standard user do not mind the `--allow-root` switch.

### DIFF
--- a/plugins/versionpress/src/Cli/VPCommandUtils.php
+++ b/plugins/versionpress/src/Cli/VPCommandUtils.php
@@ -23,6 +23,8 @@ class VPCommandUtils
             $args['color'] = null;
         }
 
+        $args['--allow-root'] = null;
+
         foreach ($args as $name => $value) {
             if (is_int($name)) { // positional argument
                 $cliCommand .= " " . ProcessUtils::escapeshellarg($value);
--- a/plugins/versionpress/src/Cli/VPCommandUtils.php
+++ b/plugins/versionpress/src/Cli/VPCommandUtils.php
@@ -23,6 +23,7 @@ class VPCommandUtils
             $args['color'] = null;
         }
 
+        // For commands that were run under root - #1049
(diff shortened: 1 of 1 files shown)


plugins/versionpress/src/Cli/VPCommandUtils.php
```diff
@@ -23,6 +23,9 @@ public static function runWpCliCommand($command, $subcommand, $args = [], $cwd =
             $args['color'] = null;
         }
 
+        // For commands that were run under root - #1049
+        $args['--allow-root'] = null;
```
borekb: This was actually not correct, fixed in https://github.com/versionpress/versionpress/pull/1272/commits/ceca06332d2d2ca6f8cfd6999e5d5526542f210c.

PR #1260, opened 2017-07-04
Temp files in ZIP

### LINKED ISSUE #1154: Unnecessary files in release 4.0.alpha1.zip
HI,

I just download latest alpha release.
https://github.com/versionpress/versionpress/releases/download/4.0-alpha1/versionpress-4.0-alpha1.zip

When i extracted the zip, i found several `.DS_Store` files in there.
Also there is a  file under `temp/_Nette.RobotLoader/_a8f7b475f99f14e7a8c92818c24898f8` which seems to contain some sensitive data. 

Are these files supposed to be in release zip ?

JanVoracek:
Resolves #1154 

We don't need the `temp` directory anymore: https://github.com/versionpress/versionpress/commit/6843b874b21fef3ea72f5e1958d40c871ecba213#diff-e8ccf25c798ea6d4068705426a027cdcR20

Security check in RequirementsChecker is now done against `versionpress-nginx.conf` (we need some non-interpreted text file).

### DIFF
--- a/plugins/versionpress/src/Utils/RequirementsChecker.php
+++ b/plugins/versionpress/src/Utils/RequirementsChecker.php
@@ -274,7 +274,7 @@ class RequirementsChecker
 
     private function tryAccessControlFiles()
     {
-        $securedUrl = site_url() . '/wp-content/plugins/versionpress/versionpress-nginx.conf';
+        $securedUrl = plugins_url('versionpress-nginx.conf', 'versionpress/versionpress.php');
         /** @noinspection PhpUsageOfSilenceOperatorInspection */
         return @file_get_contents($securedUrl) === false; // intentionally @
     }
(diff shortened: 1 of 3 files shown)


plugins/versionpress/src/Utils/RequirementsChecker.php
```diff
@@ -274,7 +274,7 @@ private function tryWrite()
 
     private function tryAccessControlFiles()
     {
-        $securedUrl = site_url() . '/wp-content/plugins/versionpress/temp/security-check.txt';
+        $securedUrl = site_url() . '/wp-content/plugins/versionpress/versionpress-nginx.conf';
```
borekb: The change is correct but maybe we have another small problem here by expecting the `wp-content/plugins` path. Maybe the [`plugins_url` function](https://developer.wordpress.org/reference/functions/plugins_url/) should be used.

PR #1314, opened 2018-01-17
Added error when re-activating VP with WP CLI

### LINKED ISSUE #1313: "wp vp activate" doesn't check if VersionPress is already active
https://github.com/versionpress/versionpress/blob/73ff46674828218c271bbcaf820c50ff4cdf8cd4/plugins/versionpress/src/Cli/vp.php#L164

Steps to reproduce:
* I installed and activated the plugin, then followed the activation procedure in the admin panel of my website. The site was successfully activated, and a .git repo with an initial commit was created in the project's root.
* Then I took a look at the CLI, and the first command was "wp vp activate". I decided to run it, for no particular reason.
* I was surprised to see that the command attempted to "activate" the site a second time.

Suggestion:

Add a check for "VersionPress::isActive()" at the start of the activate() function, and if VP is already active - print a log message and quit.

x1024:
Resolves #1313

The logging message is at the "error" level, because that's the logging level for doing "wp core download" a second time.

### DIFF
--- a/plugins/versionpress/src/Cli/vp.php
+++ b/plugins/versionpress/src/Cli/vp.php
@@ -164,6 +164,10 @@ class VPCommand extends WP_CLI_Command
     public function activate($args, $assoc_args)
     {
         global $versionPressContainer;
+        if (VersionPress::isActive()) {
+            WP_CLI::error('VersionPress is already fully activated.');
+            return;
+        }
 
         $this->checkVpRequirements($assoc_args, RequirementsChecker::SITE);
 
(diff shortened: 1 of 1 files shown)


borekb:
Nice! I also like your explanation why it's an error. The only thing I've noticed is that a file mode is changed as well:

image

That might not be intentional.

borekb:
Related to this, I'm question the term "active" as we use it:

https://github.com/versionpress/versionpress/blob/4fd75691e8d73712a1c1f7a6195c10fa17cb9596/plugins/versionpress/src/VersionPress.php#L31-L37

What's your opinion, isn't it too close to "activated" and therefore confusing?

x1024:
I updated the commit again. The permissions change was unintentional.

"isActive" is descriptive enough, and during the installation process it was made obvious what that means. I don't think there is a need to change that.

PR #1319, opened 2018-03-07
nginx redirect for /wp-json/ API endpoint

aidik:
Redirect solves the 404 error for "/wp-json/" API Endpoint when other then Plain Permalinks are used.

It is more of a hack than fix, but it can be useful until the nice(r) Permalinks are available.

### DIFF
--- a/plugins/versionpress/versionpress-nginx.conf
+++ b/plugins/versionpress/versionpress-nginx.conf
@@ -1,3 +1,7 @@
+location ~ /wp-json/.*$ {
+    rewrite ^(.*)/wp-json/(.*)$ $1/?rest_route=/$2&$args permanent;
+}
+
 location ~* /wp-content/plugins/versionpress/admin/public/.*$ {
     allow all;
 }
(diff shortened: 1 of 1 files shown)


borekb:
This is strange, WP API + permalinks + nginx usually work out of the box (we have just tested on VersionPress.com where we don't use any specific wp-json nginx rules, just the generic `try_files $uri $uri/ /index.php$args;`).

This is probably not an issue with VersionPress itself, maybe there's something wrong about your server configuration?

aidik:
@borekb You are right. It does work out of the box unless WP is installed into a subdirectory.

The scenario in which WP lives directly in http://local.dev works flawlessly. The problem emerges only at the moment WP is installed to the subdirectory e.g. http://local.dev/wordpress

It is quite possible the problem is in my server conf, I will try to figure it out.

Thank you.

borekb:
If you do, let us know please, maybe we should put a note in the documentation or something like that. Thanks!

borekb:
I'm going to close this for now, @aidik let us know if you think there are some things to improve e.g. in documentation.

Issue #1288, opened 2017-08-31
Uncommitted changes in .gitattributes file after cloning website


pavelevap:
See screenshot, `.gitattributes` content is duplicated in cloned website.

![image](https://user-images.githubusercontent.com/3008007/29915040-d7bafb60-8e3a-11e7-8fb6-0e9b2472de0f.png)

PR #1321, opened 2018-04-01
Fixed gitattributes duplicated entry on merge driver install

mkreckovic:
Resolves not detecting existing entry in gitattributes, where we end up with duplicated entry.


pavelevap:
This PR fixes #1288.

Please check Travis details:
- Remove space after opening parenthesis of function call `preg_replace()`.
- Remove blank line (74) before brace.

borekb:
@candrews I understand the pain :) I'm not convinced by this PR's implementation though. How about you opening a fresh PR with implementation that doesn't break CI and perhaps implements the logic more cleanly? (It's not obvious to my why it was necessary to flip the boolean logic and call `file_put_contents` in two distinct code paths.)

borekb:
Also, it would be good to cover this with tests.

PR #1323, opened 2018-04-01
Return paths for all plugins not only active to detect database chang…

mkreckovic:
…es while activating plugin.

### DIFF
--- a/plugins/versionpress/src/DI/DIContainer.php
+++ b/plugins/versionpress/src/DI/DIContainer.php
@@ -92,7 +92,7 @@ class DIContainer
         $dic->register(VersionPressServices::DB_SCHEMA, function () {
             global $table_prefix, $wp_db_version;
             return new DbSchemaInfo(
-                PluginDefinitionDiscovery::getPathsForActivePlugins('schema.yml'),
+                PluginDefinitionDiscovery::getPathsForPlugins('schema.yml'),
                 $table_prefix,
                 $wp_db_version
             );
@@ -107,7 +107,7 @@ class DIContainer
         });
 
         $dic->register(VersionPressServices::ACTIONSINFO_PROVIDER_ACTIVE_PLUGINS, function () {
-            return new ActionsInfoProvider(PluginDefinitionDiscovery::getPathsForActivePlugins('actions.yml'));
+            return new ActionsInfoProvider(PluginDefinitionDiscovery::getPathsForPlugins('actions.yml'));
         });
 
         $dic->register(VersionPressServices::CHANGEINFO_FACTORY, function () use ($dic) {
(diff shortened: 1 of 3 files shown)


plugins/versionpress/src/DI/DIContainer.php
```diff
@@ -92,7 +92,7 @@ public static function getConfiguredInstance()
         $dic->register(VersionPressServices::DB_SCHEMA, function () {
             global $table_prefix, $wp_db_version;
             return new DbSchemaInfo(
-                PluginDefinitionDiscovery::getPathsForActivePlugins('schema.yml'),
+                PluginDefinitionDiscovery::getPathsForPlugins('schema.yml'),
```
pavelevap: Using `get_plugins()` is a little bit problematic here, I guess. This code is [used](https://github.com/versionpress/versionpress/blob/4.0-beta/plugins/versionpress/src/Cli/vp-internal.php#L71) in `finishRestore()` function and this function is [based](https://github.com/versionpress/versionpress/blob/4.0-beta/plugins/versionpress/src/Cli/vp-internal.php#L47) on `SHORTINIT` constant, so code in `wp-settings.php` is stopped [earlier](https://github.com/WordPress/WordPress/blob/4.9.8/wp-settings.php#L130-L132) and we are [missing](https://github.com/WordPress/WordPress/blob/4.9.8/wp-settings.php#L256) definition of global variable `$wp_plugin_paths`.

Using `get_plugins()` (instead of `get_option('active_plugins')`) [leads](https://github.com/WordPress/WordPress/blob/4.9.8/wp-admin/includes/plugin.php#L291) us to `plugin_basename()` function which [uses](https://github.com/WordPress/WordPress/blob/4.9.8/wp-includes/plugin.php#L656) this global variable and suppose that `$wp_plugin_paths` is defined.

Related errors:
```
PHP Warning:  arsort() expects parameter 1 to be array, null given in C:\xampp\htdocs\vpclone\wp-includes\plugin.php on line 656
PHP Warning:  Invalid argument supplied for foreach() in C:\xampp\htdocs\vpclone\wp-includes\plugin.php on line 657
```

pavelevap:
Generally, this PR works as it should (testing manually) and database changes are tracked during plugin activation, but there are some consequences (noted above) which should be tested and examined more deeply. I am worried that mass changing `active plugins -> installed plugins` could complicate some other things.

pavelevap:
After further research I can say that problem is probably hidden somewhere else. VersionPress is [using](https://github.com/versionpress/versionpress/blob/4.0-beta/plugins/versionpress/setup-hooks.php#L117) `activated_plugin` WP hook to [run](https://github.com/versionpress/versionpress/blob/4.0-beta/plugins/versionpress/.versionpress/hooks.php#L440) `vp_plugin_changed` action which updates database schema with `refreshDbSchema()` function. But [register_activation_hook()](https://developer.wordpress.org/reference/functions/register_activation_hook/) function fires [before](https://github.com/WordPress/WordPress/blob/4.9.8/wp-admin/includes/plugin.php#L586) `activated_plugin` hook, so nothing is tracked in this situation.

I am not sure about the best solution here, maybe moving some things, for example `refreshDbSchema()` to [earlier](https://github.com/WordPress/WordPress/blob/4.9.8/wp-admin/includes/plugin.php#L571) WP hook `activate_plugin` (a while before real activation)? In this case we will need also some new method like `refreshActionsInfo()` for `ActionsInfoProvider` class (similar to `refreshDbSchema()` in `DbSchemaInfo` class) to avoid errors:

`PHP Fatal error:  Uncaught Error: Call to a member function getTags() on null in C:\xampp\htdocs\vptest\wp-content\plugins\versionpress\src\ChangeInfos\ChangeInfoFactory.php:36`

PR #1324, opened 2018-04-01
Fixed wrong SELECT and SHOW queries detection

mkreckovic:
Fixed wrong SELECT and SHOW queries detection due to leading tabs and spaces.

### DIFF
--- a/plugins/versionpress/src/Initialization/ReplacerMethods.src.php
+++ b/plugins/versionpress/src/Initialization/ReplacerMethods.src.php
@@ -103,7 +103,7 @@ class ReplacerMethods
             return $this->__wp_query($query);
         }
 
-        if (strpos(strtolower($query), 'select') === 0 || strpos(strtolower($query), 'show') === 0) {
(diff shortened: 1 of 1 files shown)


pavelevap:
Good catch! For example this [query](https://github.com/WordPress/WordPress/blob/81668f8312d2c082aeff555012a3f00f71c1e202/wp-includes/comment.php#L378) from WP core is also affected.

Originally I thought that this is handled by WordPress by default, but it is done later :-( VersionPress is using custom method `query()` before using `__wp_query()` based on original function which handles queries with [strip_invalid_text_from_query()](https://developer.wordpress.org/reference/classes/wpdb/strip_invalid_text_from_query/) method. Maybe we could take some more inspiration from it, they are using `ltrim()` and also `DESCRIBE`, `DESC` and `EXPLAIN` queries could be ignored (not very frequent).

PR works for me, thank you @mkreckovic!

candrews:
@borekb can you please take a look and perhaps merge this one?

borekb:
It looks good to me, sorry for not getting to this earlier. Just to confirm, @candrews, you find the code correct, right?

borekb:
Thanks for confirming. I'm going to merge the PR then – props to @mkreckovic!

PR #1325, opened 2018-04-01
Fixed regex not including multiline strings while serializing

mkreckovic:
Fixed regex not including multiline strings while serializing, if string starts with <some_string>


pavelevap:
Hi @mkreckovic,

I tried to test your fix, but there is no change for me. Serialized multiline strings are missing line breaks during `.ini` conversion which is probably some bug. Also strings with leading `<some_string>` are processed (but without line breaks). Do you have any particular example to test?

Some tests will be also needed.

PR #1326, opened 2018-04-01
Added support for nested shortcodes

mkreckovic:
If shortcode has content betwen shortcode open and closing markup, it is ignored.

### DIFF
--- a/plugins/versionpress/src/Database/ShortcodesReplacer.php
+++ b/plugins/versionpress/src/Database/ShortcodesReplacer.php
@@ -142,6 +142,15 @@ class ShortcodesReplacer
             }
 
             $shortcodeTag = $m[2];
+            $shortcodeSelfClosing = $m[4];
+            $shortcodeContent = $m[5];
+            if (!empty($shortcodeContent)) {
+                if ($idProvider[1] == 'getIdByVpid') {
+                    $shortcodeContent = $this->restoreShortcodes($shortcodeContent);
+                } elseif ($idProvider[1] == 'getVpidByEntityNameAndId') {
+                    $shortcodeContent = $this->replaceShortcodes($shortcodeContent);
+                }
+            }
             $shortcodeInfo = $shortcodesInfo->getShortcodeInfo($shortcodeTag);
             $attributes = shortcode_parse_atts($m[3]) ?: [];
 
@@ -155,7 +164,7 @@ class ShortcodesReplacer
                 }
             }
 
-            return WordPressMissingFunctions::renderShortcode($shortcodeTag, $attributes);
+            return WordPressMissingFunctions::renderShortcode($shortcodeTag, $attributes, $shortcodeSelfClosing, $shortcodeContent);
         };
     }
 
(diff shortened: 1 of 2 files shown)


pavelevap:
Great, you are right, I tested nested shortcodes without PR and received following results:

:heavy_check_mark:  `[outer][test id="12"][/outer]`, `[outer1][outer2][test id="12"][/outer2][/outer1]`

:exclamation: `[test id="12"][inner][/test]`, `[test id="12"][inner1][inner2][/inner1][/test]`

Then I applied your PR and problem was fixed. But, it caused another issue. Simple shortcode `[test id="12"]` (and also related nested shortcodes) is transformed to `[test id="12"][/test]` (changing self-closing shortcode to enclosing shortcode).

Also tests should be part of this PR.

Another observation from testing (not related to this PR): When using `[test id=12]`, `[test id="12"]` or `[test id='12']`, VersionPress transforms everything into double quotes version. I am not sure if it can cause any troubles. @JanVoracek

mkreckovic:
@pavelevap Wordpress built-in function `get_shortcode_regex` returns as much information that it is possible to make only three types of shortcodes:
* `[%s %s]%s[/%s]`
* `[%s %s][/%s]`
* `[%s %s /]`
So, we have a situation when we convert `[%s %s]` to `[%s %s /]` shortcode format, but these are equivalents so it's not a problem.

From [Wordpress](https://codex.wordpress.org/Shortcode_API):

> The parser supports xhtml-style closing shortcodes like "[myshortcode /]", but this is optional.

> The self-closing marker, a single forward slash, is optional. Space before the marker is optional. Spaces are not allowed after the marker.

As far as the quotes are concerned, it is also not a problem:

> Shortcode macros may use single or double quotes for attribute values, or omit them entirely if the attribute value does not contain spaces. [myshortcode foo='123' bar=456] is equivalent to [myshortcode foo="123" bar="456"].

borekb:
@mkreckovic The code begs for updating [`ShortcodesReplacerTest`](https://github.com/versionpress/versionpress/blob/master/plugins/versionpress/tests/Unit/ShortcodesReplacerTest.php) as well :) Here are our [docs on running tests](https://github.com/versionpress/versionpress/blob/master/docs/Dev-Setup.md#testing).

And thanks for the contribution!

pavelevap:
@mkreckovic: Yes, you are right, but my concern is VersionPress usage. You can try for example default gallery shortcode. Insert any gallery into post and you will have `[gallery ids="12"]` in post content (database). But related `.ini` file will contain `[gallery ids="VPID"][/gallery]` and this should be consistent, I guess. When you use undo or rollback, post content will be updated from related `.ini` file and it can lead to conflicts between environments. WordPress (shortcodes) will work without problems, but I am not sure about VersionPress.

pavelevap:
All unit tests for shortcodes are failing with this PR, because of this change:

image

I still think that we should not change post content in this way (see my previous comment).

pavelevap:
Also Synchronizer tests are failing ([synchronizerRestoresShortcodes](https://github.com/versionpress/versionpress/blob/0ecf1d2a49d0e7cf63df88e23807c42d0fec104d/plugins/versionpress/tests/SynchronizerTests/PostsSynchronizerTest.php#L269)):

image

Issue #1304, opened 2017-12-12
Missing custom schemas on cloned site


pavelevap:
- When you clone a site, folder `wp-content/.versionpress/` is missing on cloned site.

On original site I defined postmeta `test` as entity related to `post`. I added value `33` and it was correctly transformed to `337025C11E10459880A239700C41A057` file in VersionPress. But on a cloned site is value `337025C11E10459880A239700C41A057` saved into database for postmeta `test`.

PR #1327, opened 2018-04-01
Fixed gitignore template to include plugin definitions

mkreckovic:
Fixed gitignore template to include plugin definitions and added some new entries

### DIFF
--- a/plugins/versionpress/src/Initialization/.gitignore.tpl
+++ b/plugins/versionpress/src/Initialization/.gitignore.tpl
@@ -8,17 +8,27 @@
 versionpress.maintenance
 /.htaccess
 /web.config
+.well-known/
 
 {{wp-content}}/*
 !{{wp-content}}/db.php
 !{{wp-content}}/index.php
 !{{wp-plugins}}/
 {{wp-plugins}}/versionpress/
+!{{wp-content}}/.versionpress/
 !{{wp-content}}/mu-plugins/
 !{{wp-content}}/themes/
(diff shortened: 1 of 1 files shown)


plugins/versionpress/src/Initialization/.gitignore.tpl
```diff
@@ -8,19 +8,29 @@
...
+!{{wp-content}}/.versionpress/
 !{{wp-content}}/mu-plugins/
 !{{wp-content}}/themes/
 !{{wp-content}}/languages/
 !{{wp-content}}/uploads/
 !{{wp-content}}/vpdb/
 
 
+#------------------------
+#  Cache files
```
pavelevap: I am not sure if this should be added into `.gitignore` template or handled by rules for particular plugins. But it is probably quick win for now and helpfull for users of caching plugins. Folder `/uploads/et_temp` is related to Divi builder?

pavelevap:
Great, this PR fixes #1304.

borekb:
Thanks for the PR, I've updated the description to follow our [standard template](https://github.com/versionpress/versionpress/blob/2071052ad33a782e96ca206ccca6891e52bc614f/.github/PULL_REQUEST_TEMPLATE.md).

What I'd suggest is to keep this PR to the actual fix. For example, I don't think we should be ignoring IDE project files by default – that is highly subjective and each team have their preference. Caching rules are another example pointed out by @pavelevap. Generally, it would be easier if this PR did one thing only.

PR #1340, opened 2018-06-29
Docs for VP_WP_CLI_BINARY config option

borekb:
Introduced in #1187, we forgot to document it.

I've also fixed the Markdown formatting in `configuration.md`.


borekb:
It's a good point that the info about `wp-config.common.php` should probably be extracted to its own docs page; it's a standalone concept that exists even if there were no VersionPress config constants.

I'll keep this PR about `VP_WP_CLI_BINARY` only though; let's revisit that in the future.

PR #1352, opened 2018-09-13
openid_nonces vary by server so should be ignored

candrews:
openid_nonces are tied to the Wordpress site URL, so they shouldn't be handled by VersionPress

Resolves #


pavelevap:
Hi @candrews,

`openid_nonces` option is not a part of WordPress core? It is probably related to some plugin and you can manually add support for it, please see our [docs](https://docs.versionpress.net/en/developer/plugin-support/). Online repository for plugin definitions is planned, see #1243.

candrews:
My apologies, I thought it was part of core. It's actually from https://wordpress.org/plugins/openid/

Sorry about that!

Issue #1350, opened 2018-08-28
Plugin support: how to ignore a "serialized object" entity


beabri:
Hi all,

inside the file `schema.yml` of wordpress-seo plugin I put some options to let them be ignored during the synchronization, for example

```
option:
  ignored-entities:
    - 'option_name: wpseo_tracking_last_request'
    - 'option_name: wpseo_sitemap_*_cache_validator'
    - 'option_name: wpseo_sitemap_cache_validator_global'
    - 'option_name: recently_activated'
```

This way, synchronization ignores the entire option_value.

I need to insert a option which is stored in DB this way:

![](https://files.gitter.im/versionpress/versionpress/4KF5/image.png)

The option_value is a serialized object which, unserialized, is an array like this:

```
Array
(
    [ms_defaults_set] => 
    [version] => 8.0
    [disableadvanced_meta] => 1
    [onpage_indexability] => 1
    [baiduverify] => 
    [googleverify] => 
    [msverify] => 
    [yandexverify] => 
    [site_type] => 
    [has_multiple_authors] => 
    [environment_type] => 
    [content_analysis_active] => 1
    [keyword_analysis_active] => 1
    [enable_admin_bar_menu] => 1
    [enable_cornerstone_content] => 1
    [enable_xml_sitemap] => 1
    [enable_text_link_counter] => 1
    [show_onboarding_notice] => 
    ----> [first_activated_on] => 1533804994
)
```

I want to insert just the "first_activated_on" value, not all the option_value, to be ignored during synchronization. Is there a way to do it?

I tried this:

`'option_name@option_value: wp_seo[/\d+/]["first_activated_on"]'`

But without success.

Thank you in advance.

borekb:
Thanks for creating this issue, we indeed do not support this scenario yet and I agree it would be useful to have this for 4.0 final.

PR #1354, opened 2018-09-19
Ignore user_meta community-events-location

candrews:
community-events-location records the user's IP upon login; that's not useful information to store in version control.

Resolves #

### DIFF
--- a/plugins/versionpress/.versionpress/schema.yml
+++ b/plugins/versionpress/.versionpress/schema.yml
@@ -104,6 +104,7 @@ usermeta:
     - 'meta_key: nav_menu_recently_edited'
     - 'meta_key: wporg_favorites'
     - 'meta_key: *dashboard_quick_press_last_post_id'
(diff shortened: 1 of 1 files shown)


borekb:
Hi @candrews, thanks for the PR. It looks good to me, I've asked @pavelevap for one final confirmation and then we can merge this. Thanks!

pavelevap:
Hi @candrews,

I checked `community-events-location` usermeta, but it is not used only for IP address.
Users can also manually select their own city and ignore IP address, please see screenshot. In this case we should preserve this information, I guess.
image
I agree that we can ignore IP address, but it depends on another related issue: #1350 (ignoring only part of serialized entity).

borekb:
Nice find. So I guess it would be best to wait for #1350 to be implemented and then re-do the schema change using the update API / syntax in a fresh PR. The caveat here is that #1350 is not easy to implement, unfortunately, so I'm not sure when we'll have the bandwidth to do it.

PR #1360, opened 2018-11-12
1359 :: Synchronizer : Add Error Log

### LINKED ISSUE #1359: Synchronizer fails silently when encountering unexpected vp ids.
Hello VP team!

Thanks for making such an awesome plugin/tool. Database synchronization between environments is always a challenging problem and your plugin makes this much easier! 

We recently came across an a bug that really had us stumped and thought that we would share our findings to hopefully help others in the same spot and possibly advocate for some code updates. This was the symptom:
  

>   When on our "production" branch everything in the world was fine/happy. We were able to build to local environments from a clean install (using docker) and switch to the branch carrying over any/all changes using `wp vp apply-changes`.
> 
>   When on our staging branch on the other hand the world was such a fun place. The main issue we were seeing is that the `wp_term_relationships` table was all wrong (it was empty). It appeared that from a fresh installation on our local machines that the table was never populated and when changing branches over to staging that no changes were applied. In fact if we ran `wp vp apply-changes` the `wp_term_relationships` table would get completely emptied out :-(
> 
>   Changes made directly through the admin while on the staging branch seemed to persist fine, it was just that initial build and then pulling in changes that seemed to cause issue. 
> 

After lots of exploration and hair pulling we were able to find the route of the issue:

  1. About a month back we purged our site (though code) of all tags and added a more selective group of them.
    - Due to the dependence of WP `$post->ID`s we had to this on production (since that is the content team supplied the mappings from)
    - In developing the solution for the tag update the tag deleting script was run on staging _as well as_ production. **I think this is where our troubles began**
  2. We continued with active development on both staging and production. Fast forward to about a week ago and our content managers started expressing concerns about the missing taxonomy relations on stage.
  3. We were able to isolate the issue to the missing content in the `wp_term_relationships` table.
  4. We were able to trace the code that should be filling that table to the `Synchronizer` class (in `Synchronizer.php`).
  5. In there we discovered that the `idMapContainsAllVpIds` function was very strict in its requiring all IDs to map properly to and that when a "bad" ID was encountered that the code failed silently. 
    - I believe that the reason for this very strict code is due to the assumed happy path use of the `$idMap` variable in the relational mapping method `fixMnReferences` which allows for this sweet block of PHP deliciousness:
```
      $valuesForInsert = array_map(function ($relation) use ($idMap) {
          $sourceId = $idMap[$relation['vp_id']];
          $targetId = $idMap[$relation['referenced_vp_id']];
          return "($sourceId, $targetId)";
      }, $relations);
``` 
_Notice the lack of checking `if( !empty($idMap[$relation['vp_id']]) )` or similar._
  6. Throwing a log in place of the `return false` in `idMapContainsAllVpIds` revealed that we had 10 invalid vp_ids which we were able to manually reconcile.
  7. With the code back on the "happy path" everything was right in the world again and VP worked its magic as expected.

Obviously this is a fringe case and we would assume that 99% of users will not run into this sort of issue. With that said we were wondering if there is value in adding some sort of logging when a bad ID is encountered (instead of failing silently) and/or perhaps just skipping that one ID (or group of IDs)?

Thanks again for the great plugin and hopefully this post helps the sanity of some other developer out there.

ben-caplan:
- Adds error log to prevent code from failing silently.

Resolves #1359

### DIFF
--- a/plugins/versionpress/src/Synchronizers/Synchronizer.php
+++ b/plugins/versionpress/src/Synchronizers/Synchronizer.php
@@ -633,6 +633,9 @@ class Synchronizer
     {
         foreach ($vpIds as $vpId) {
             if (!isset($idMap[$vpId])) {
+                error_log(
+                    '[VERSION PRESS]: Error! An invalid $vpId was encountered (' . $vpId . '). As such *ALL* relationships are being skipped, this means everything from the "wp_term_relationships" table (associations of all taxonomies to all post types, custom menus, etc). Please manually resolve this unknown ID and re-run "wp vp apply-changes" (or comparable action).'
+                );
                 return false;
             }
         }


borekb:
Hi @ben-caplan, thanks for the PR.

We've looked into this closer with @JanVoracek today and it turns out that it won't be that simple. For reference, the logging is added here, to the `idMapContainsAllVpIds` function:

https://github.com/versionpress/versionpress/blob/f20835148e0057806ce2414a05c97a498b2565e8/plugins/versionpress/src/Synchronizers/Synchronizer.php#L481-L485

This function is called from `fixMnReferences` which is called twice from the main `synchronize` method. The first call looks like this:

https://github.com/versionpress/versionpress/blob/f20835148e0057806ce2414a05c97a498b2565e8/plugins/versionpress/src/Synchronizers/Synchronizer.php#L146-L153

This code tries to fix M:N references but if it can't, it's fine – it will just add it to the list of tasks to execute in the second pass. The reason is that in some cases, not all entities are yet in the database, for example, when a post with a tag is being synchronized. In that case, it is expected not to find an ID in the database so the proposed code would produce warnings even when it shouldn't.

In the second pass, the function is called again here, which would be the right place to add logging:

https://github.com/versionpress/versionpress/blob/f20835148e0057806ce2414a05c97a498b2565e8/plugins/versionpress/src/Synchronizers/Synchronizer.php#L168-L170

However, I'm not sure we should be adding a one-off message like this. Maybe a better approach would be to remove the true/false nature of `fixMnReferences`, always postpone it to the second pass and let it fail loudly, similarly to the other `fixSomething` methods. What do you think?

ben-caplan:
Hi @borekb, 

Thanks for having a look at my PR and for your feedback! What you are saying make sense to me. Would the idea then be to remove the use of `idMapContainsAllVpIds` in addition to allowing/forcing the double loop (by removing the  true/false nature of `fixMnReferences`)? I assume that would through unwanted errors in the first loop you described. If you and/or @JanVoracek have a clear picture of what the code would look like to allow for failing loudly but only on the second loop, that sounds like a solid option.

One other thing that occurred to me, assuming that you are ok with the general idea of specific logging messages, is that we could just put something like this before the error log (pseudo code): 
```
if( in_array(self::SYNCHRONIZE_MN_REFERENCES, $tasks) ){
  error_log('...');
}
``` 
The idea being to make sure that the code fires as part of the second loop. With that said I am open to either approach so long as there is some sort of hint/clue as why the expected result is not reached. The more useful meaningful message would be nice and easier to troubleshoot, but I will take whatever I can get 😄

Let me know what you think and I am happy to help out in any way that I can. Thanks again for your time and help with this!

borekb:
> Would the idea then be to remove the use of `idMapContainsAllVpIds` in addition to allowing/forcing the double loop (by removing the true/false nature of `fixMnReferences`)?

Yes, that would be the idea. If you could try that (and ideally [run tests](https://docs.versionpress.net/en/developer/dev-setup/#testing) after that), that would be great.

PR #1375, opened 2019-02-25
require_once the Wordpress includes

### LINKED ISSUE #1374: restore-site doesn't work with Wordpress 5.1
When running `wp vp restore-site --yes --siteurl="${SITE_URL}" --require=wp-content/plugins/versionpress/src/Cli/vp.php` with `wp-config.php` present against an empty database (with no database tables), it fails with this output:
```
Success: Database tables created
PHP Fatal error:  Cannot redeclare wptexturize() (previously declared in /var/www/html/wp-includes/formatting.php:51) in /var/www/html/wp-includes/formatting.php on line 315
Error: Could not finish site restore
```

It works fine with Wordpress 5.0.

candrews:
Using require can cause the files being included once to result in an error as it does on Wordpress 5.1.

Resolves #1374


borekb:
Thank you!

PR #1377, opened 2019-03-04
Upgrade to PHP_CodeSniffer and fix newly identified issues

candrews:
Upgrade to the latest PHP_CodeSniffer (3.4.0). `( cd plugins/versionpress && ./vendor/bin/phpcs --standard=ruleset.xml )` then fails due to formatting issues; so fix those issues.


borekb:
Looks good, thank you!

PR #1378, opened 2019-03-04
Use php7.2 in Travis to be consistent with the docker images

candrews:
The docker images all use PHP 7.2, so for consistency at least, Travis should also use PHP 7.2.


borekb:
👍, thanks.

PR #1380, opened 2019-03-04
/var/www/html in wordpress-for-tests should not be a volume

candrews:
This directory is used for running tests; it is entirely ephemeral.

Also, having it as a volume causes the mkdir and chown in
./dev-env/wordpress-cli-image/Dockerfile
to not take effect (since the volume is mounted over
/var/www/html), resulting in the tests not being able to
write to /var/www/html since its not owned by www-data,
resulting in tests failing with this error:
```
Exception: Error executing cmd 'php '/tmp/wp-cli-latest-stable.phar' core download --path='/var/www/html/wptest' --version='4.9' --force --locale='en_US'' from working directory '/var/www/html/wptest':
Error: '/var/www/html/wptest/' is not writable by current user.
 in /opt/versionpress/tests/Automation/WpAutomation.php:621
```

Resolves #

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,8 +18,6 @@ services:
     image: versionpress/wordpress:php7.2-apache
     ports:
       - "80:80"
-    volumes:
-      - ./dev-env/wp-for-tests:/var/www/html:z
     links:
       - mysql-for-tests:mysql
     working_dir: /var/www/html/wptest
@@ -73,7 +71,6 @@ services:
       PHP_IDE_CONFIG: serverName=VersionPress-tests
     volumes:
       # !!! This must be kept in sync with wordpress-cli-image/Dockerfile
-      - ./dev-env/wp-for-tests:/var/www/html:z
       - ./dev-env/test-logs:/var/opt/versionpress/logs:z
       - ./plugins/versionpress:/opt/versionpress:ro,z
       - ./ext-libs:/opt/ext-libs:ro,z


borekb:
I forgot the details but the idea was that if your tests fail, you can inspect the folder as it's mounted to `dev-env/wp-for-tests`. What would the workflow be after the proposed change?

candrews:
You could exec into the docker container and look that way.

borekb:
Do you know why we don't have a similar problem with other mounts from `dev-env`? The goal is to have various folders mapped to a host OS filesystem which is much more developer-friendly than having to inspect containers via `docker exec`. I've also never seen the permission issue myself, could it be related to your OS?

Don't get me wrong, your reasoning sounds solid to me but I care about the experience of troubleshooting failed tests.

candrews:
> could it be related to your OS?

Travis CI experiencing the exact same error as I do on my Fedora system, so if it's OS dependent, it's probably for all Linux distros.

> I care about the experience of troubleshooting failed tests.
I do too :) But right now, we have no CI test running and tests are failing.

Do you have other thoughts as to how we can solve this problem? It looks like you worked on a closely related issue in 87769bf2e

borekb:
Yes, I was quite familiar with many of the details a year ago, more than I would have liked :) I would have to research it again though, which I'm happy to do after I return from holiday later this week.

candrews:
Superceded by #1388

PR #1379, opened 2019-03-04
Make sure the test-logs directory exists

candrews:
The directory must exist. If it doesn't, tests fail to run with this output:
```
$ npm run tests:full

> versionpress-dev-workspace@1.0.0 tests:full /home/candrews/projects/versionpress
> docker-compose run --rm tests-with-wordpress ../vendor/bin/phpunit -c phpunit.xml

Starting versionpress_mysql-for-tests_1 ... done
Starting versionpress_selenium-hub_1    ... done
Starting versionpress_wordpress-for-tests_1 ... done
RuntimeException: Logging directory not found or is not absolute path. in /opt/versionpress/vendor/tracy/tracy/src/Tracy/Debugger.php:162
Stack trace:
#0 /opt/versionpress/tests/phpunit-bootstrap.php(13): Tracy\Debugger::enable(false, '/var/opt/versio...')
#1 /opt/versionpress/vendor/phpunit/phpunit/src/Util/Fileloader.php(52): include_once('/opt/versionpre...')
#2 /opt/versionpress/vendor/phpunit/phpunit/src/Util/Fileloader.php(36): PHPUnit_Util_Fileloader::load('/opt/versionpre...')
#3 /opt/versionpress/vendor/phpunit/phpunit/src/TextUI/Command.php(914): PHPUnit_Util_Fileloader::checkAndLoad('/opt/versionpre...')
#4 /opt/versionpress/vendor/phpunit/phpunit/src/TextUI/Command.php(739): PHPUnit_TextUI_Command->handleBootstrap('/opt/versionpre...')
#5 /opt/versionpress/vendor/phpunit/phpunit/src/TextUI/Command.php(127): PHPUnit_TextUI_Command->handleArguments(Array)
#6 /opt/versionpress/vendor/phpunit/phpunit/src/TextUI/Command.php(116): PHPUnit_TextUI_Command->run(Array, true)
#7 /opt/versionpress/vendor/phpunit/phpunit/phpunit(52): PHPUnit_TextUI_Command::main()
#8 {main}
Unable to log error: Directory is not specified.
npm ERR! code ELIFECYCLE
npm ERR! errno 255
npm ERR! versionpress-dev-workspace@1.0.0 tests:full: `docker-compose run --rm tests-with-wordpress ../vendor/bin/phpunit -c phpunit.xml`
npm ERR! Exit status 255
npm ERR! 
npm ERR! Failed at the versionpress-dev-workspace@1.0.0 tests:full script.
[...]

PR #1381, opened 2019-03-04
Run tests in Travis

candrews:
Have Travis run all tests:
  - npm run lint:markdown
  - npm run build-images
  - npm run tests:full
  - ( cd plugins/versionpress && ./vendor/bin/phpcs --standard=ruleset.xml )

With this improvement, tests are automatically run so no one has to manually do so. Tests are also run for all branches and PRs allowing for test passage to be a quality gate.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,8 +12,6 @@ cache:
     - $HOME/.composer/cache
 
 before_install:
-  - git config --global github.accesstoken $GITHUB_OAUTH_TOKEN
-  - composer config -g github-oauth.github.com $GITHUB_OAUTH_TOKEN --no-interaction
 
 script:
     cd plugins/versionpress && composer install && ./vendor/bin/phpcs --standard=ruleset.xml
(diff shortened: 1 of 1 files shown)


candrews:
Note that tests currently fail; #1376 #1379 #1380 all work towards making these tests pass.

borekb:
This is awesome ❤️!

I'm inclined to merge this before the tests pass, do you agree?

candrews:
Absolutely!

borekb:
I wonder if it's useful to run `build-images` on Travis, what do you think @candrews?

candrews:
> I wonder if it's useful to run `build-images` on Travis

I think it is.

If Travis doesn't build the images, then Travis can't test changes to the images. For example, if you modify a Dockerfile, and Travis isn't building images, then Travis won't be able to run tests with that change in place. This means docker changes can only be tested locally or after the image is pushed.

borekb:
👍, you're right. It's a bit wasteful most of the time but we could add something like [this](https://stackoverflow.com/questions/47174711/perform-docker-build-only-if-dockerfile-is-changed) in the future.

PR #1385, opened 2019-03-06
Using "\" to escape "-" in tr isn't standard so move "-" to the end to avoid having to escape it

### LINKED ISSUE #1384: Merge driver doesn't work on Debian-based systems
We currently maintain two implementations of a Git merge driver, [`ini-merge.sh`](https://github.com/versionpress/versionpress/blob/b14fe014360ae36597ac2fd18effa801de566f31/plugins/versionpress/src/Git/merge-drivers/ini-merge.sh) and [`ini-merge.php`](https://github.com/versionpress/versionpress/blob/b14fe014360ae36597ac2fd18effa801de566f31/plugins/versionpress/src/Git/merge-drivers/ini-merge.php). The `.sh` one is used on Linux and macOS systems and we tried to make it compatible with various Bash and `sh` implementations, see e.g. https://github.com/versionpress/versionpress/commit/a6e05c032c5919c81fb4102cf00535c7e773e53e. However, it seems that we found one that doesn't work, through a failed [`CloneMergeTest#dateModifiedMergesAutomatically`](https://github.com/versionpress/versionpress/blob/e2209836d30889727991365fde2c983d0121e5a1/plugins/versionpress/tests/Workflow/CloneMergeTest.php#L88) test.

(Note: for a while, we thought that the problem is in the test, see PR #1376 and an issue #1382 but it's really a Bash / sh incompatibility.)

This test runs on our [WordPress Docker image](https://github.com/versionpress/versionpress/blob/b14fe014360ae36597ac2fd18effa801de566f31/dev-env/wordpress-image/Dockerfile) which is based on [wordpress:php7.2-apache](https://github.com/docker-library/wordpress/blob/b8bfd453a65ce574982d328103aaa2850997cede/php7.2/apache/Dockerfile) which is based on **debian:stretch-slim**. For comparison, the merge driver is also tested by `MergeDriverTest`, specifically by [`mergedDatesWithoutConflict`](https://github.com/versionpress/versionpress/blob/b14fe014360ae36597ac2fd18effa801de566f31/plugins/versionpress/tests/GitRepositoryTests/MergeDriverTest.php#L133) and all is fine there. The only substantial difference is that these tests run on the [`versionpress/wordpress:cli` image](https://github.com/versionpress/versionpress/blob/b14fe014360ae36597ac2fd18effa801de566f31/dev-env/wordpress-cli-image/Dockerfile) which is based on [wordpress:cli](https://github.com/docker-library/wordpress/blob/b8bfd453a65ce574982d328103aaa2850997cede/php7.2/cli/Dockerfile) which is based on **Alpine Linux**.

I'm not quite sure if the "stretch-slim" image is heavily customized or if the bug affects all Debian-based distributions but it's quite bad anyway as it would affect all users of the official WordPress Docker image.

The best way to fix this would be to re-implement the merge drive in Go as suggested in #736 and drop both the Shell and PHP implementations. It's a relatively ugly and fragile code and we don't want to maintain it twice.

I'll keep this issue open until the Go implementation is ready.

candrews:
Fixes #1384

### DIFF
--- a/plugins/versionpress/src/Git/merge-drivers/ini-merge.sh
+++ b/plugins/versionpress/src/Git/merge-drivers/ini-merge.sh
@@ -24,8 +24,8 @@ do
     B_DATE_STR=$(printf "%s" "$B_DATE_STR" | tr -d '\r\n')
 
     # Transform them to Numbers
-    A_DATE_NUM=$(printf "%s" "$A_DATE_STR" | tr -d "\-: ")
-    B_DATE_NUM=$(printf "%s" "$B_DATE_STR" | tr -d "\-: ")
+    A_DATE_NUM=$(printf "%s" "$A_DATE_STR" | tr -d ": -")
+    B_DATE_NUM=$(printf "%s" "$B_DATE_STR" | tr -d ": -")
 
     # Compare and make both values same
     if [ "$A_DATE_NUM" -lt "$B_DATE_NUM" ]; then


candrews:
@borekb What do you think?

borekb:
I'm trying to run tests for it, and doing a full suite was not the best idea :) It's been almost an hour an a half since it started.

Do you have a couple of examples of how the replacements actually look, before and after the fix?

candrews:
> Do you have a couple of examples of how the replacements actually look, before and after the fix?

test
```
Begin
post_modified = "2013-03-03 13:13:13"
End
```

test2
```
Begin
post_modified = "2014-03-03 13:13:13"
End
```

test3
```
Begin
post_modified = "2015-03-03 13:13:13"
End
```

test2 is the file that will change.

Testing done on by running ini-merge.sh on `versionpress/wordpress:cli` (which uses Alpine and therefore busybox tr, not gnu tr).

Before this change, test2:
```
Begin
post_modified = "2014-03-03 13:13:13"
End
```
Also, this error is output to stdout: `sh: 2014-03-0313:13:13: bad number`

With this change, test2:
```
Begin
post_modified = "2015-03-03 13:13:13"
End
```
and no error is output to stdout.

borekb:
Awesome! When the tests finish (I'm now running only `CloneMergeTest` and it still takes ages!), I'm going to merge this.

Really glad you managed to fix this. We already have some friends working on the Go implementation but it might take a while so this is greatly appreciated 👍.

pavelevap:
I tried only `dateModifiedMergesAutomatically` test and it works for me :tada:

borekb:
For me too. Thanks, @candrews!

pavelevap:
I am not sure if it is related, but running similar `mergedDatesWithoutConflict` test leads me to following Warning:

`Warning: chmod(): Read-only file system in /opt/versionpress/src/Git/MergeDriverInstaller.php on line 97`

https://github.com/versionpress/versionpress/blob/e2209836d30889727991365fde2c983d0121e5a1/plugins/versionpress/src/Git/MergeDriverInstaller.php#L97

Oh, now I see, that I removed `@` yesterday during tests, so probably nothing important.

Issue #1383, opened 2019-03-05
Testing is problematic on Linux


borekb:
As [mentioned](https://gitter.im/versionpress/versionpress?at=5c7d56b1d2d62067b741fbd2) by @candrews on Gitter, running tests is currently problematic on Linux (Craig uses Fedora if that's important). The core VersionPress team only uses macOS or Windows so we're not really sure what's going on on Linux but it already lead to PRs such as #1371, #1379 or #1380 which sometimes feel a bit arbitrary to me. Are we doing something fundamentally wrong about permissions, filesystem or something like that?

We'd appreciate help from those of you with Linux / Docker experience. Our dockerized dev stack was built primarily in these two pull requests:

- https://github.com/versionpress/versionpress/pull/1227 (merged July 2017)
- https://github.com/versionpress/versionpress/pull/1329 (merged June 2018)

@candrews, could you possibly involve someone from Isobar?

borekb:
@candrews could you please try this?

```
docker run --rm -it -v $PWD/dev-env/testfolder:/var/www/html versionpress/wordpress:cli /bin/sh
```

Does this run successfully and create `dev-env/testfolder` on your host system? From within the container, is `/var/www/html` owned by www-data? Thanks.

borekb:
This is an issue from the official WordPress image repository talking about Fedora and FS permissions, seems possibly related to me: https://github.com/docker-library/wordpress/issues/132.

candrews:
Due to SELinux, the command should use `z` or `Z`, so I actually ran:
`docker run --rm -it -v $PWD/dev-env/testfolder:/var/www/html:z versionpress/wordpress:cli /bin/sh`

> Does this run successfully and create `dev-env/testfolder` on your host system?

Yes.
From the host:
```
$ ls dev-env/testfolder/ -la
total 8
drwxr-xr-x. 2 root root 4096 Mar  5 16:20 .
drwxr-xr-x. 3 root root 4096 Mar  5 16:20 ..
```

> From within the container, is `/var/www/html` owned by www-data? Thanks.

No. from within the container:
```
$ ls -la /var/www/
total 20
drwxr-xr-x    1 root     root          4096 Mar  2 02:57 .
drwxr-xr-x    1 root     root          4096 Mar  2 02:57 ..
drwxr-xr-x    2 www-data www-data      4096 Mar  2 02:57 .wp-cli
drwxr-xr-x    2 root     root          4096 Mar  5 21:20 html
```

[...]

PR #1388, opened 2019-03-06
Run cli as user 33:33 to match www-data from versionpress/wordpress:php7.2-apache

candrews:
On Debian, www-data:www-data is 33:33
On Alpine (which is what cli is based on), www-data:www-data is 82:82
Therefore, when Alpine and Debian interact, one has to change to match the other.

See https://github.com/docker-library/wordpress/issues/256 for details

Makes progress on #1383, Travis now gets further.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -64,6 +64,7 @@ services:
       - wpcli-cache:/var/www/.wp-cli:z
     working_dir: /opt/versionpress/tests
     command: ../vendor/bin/phpunit --verbose --colors -c phpunit.xml --testdox-text /var/opt/versionpress/logs/testdox.txt
+    user: "33:33"
 
   tests-with-wordpress:
     image: versionpress/wordpress:cli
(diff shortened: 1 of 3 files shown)


borekb:
That looks good. Does that impact non-Linux OS's in some way? I'll try to run some tests on macOS shortly.

candrews:
I just added a commit to not do sudo, chown, and chgrp. That's all unnecessary since the tests already run as the right user. And doing those operations undoes the `user 33:33` change from the other commit in the PR, causing failures.

borekb:
@candrews I'll go get some sleep but if you want me to run something tomorrow morning, let me know. I'm looking forward to seeing all tests pass again :)

candrews:
With these changes, it's down to 19 errors and 16 skipped. So that's a nice improvement!

borekb:
Thanks!

borekb:
Follow-up PR to this is #1397 which also sets the correct user for the mount points inside the Dockerfile. That resolves issues around test-logs and similar.

PR #1391, opened 2019-03-09
Add go version merge driver

### LINKED ISSUE #736: Implement custom merge driver in Go
In #588 we created merge driver in Bash for Linux/Mac and in PHP for Windows. On Windows, there is a significant difference between time required to merge without custom merge driver (9.5 seconds) and with custom merge driver in PHP (110 seconds) - see [full results](https://github.com/versionpress/versionpress/issues/588#issuecomment-185712454). Because of the difference, we should try to implement the driver in some language with smaller overhead.

The key issue is that we need an implementation we can compile to the native code for all platforms – without any tools on the user side. Thus, we probably cannot use languages like Python, Perl, etc.

We thought about C or C++ but neither of us is fluent in them. Thus, maybe Go or Lua will be better :)

abtris:
Resolves #736

This is naive implementation php version in go lang.

I try tested using my test repository but getting this error:

```
2019/03/09 07:06:28 Command finished with error: exec: "git merge-file -L mine -L base -L theirs .merge_file_TO39EH .merge_file_OCO0Kc .merge_file_TXE15j": executable file not found in $PATH
```
I added lookPath part if is problem with `git` or binary `ini-merge` but don't find problem.

### DIFF
--- a/plugins/versionpress/src/Git/merge-drivers/ini-merge.go
+++ b/plugins/versionpress/src/Git/merge-drivers/ini-merge.go
@@ -1,0 +1,147 @@
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+	"os/exec"
+	"regexp"
+	"strings"
+	"time"
+)
+
+func main() {
+	// File names passed via cli arguments
+	checkArgs("<current-name> <base-name> <other-name>")
+	log.Printf("%v", os.Args)
+	O := os.Args[1]
+	A := os.Args[2]
+	B := os.Args[3]
+	// Dates fields to merge
+	dates := []string{"post_modified", "post_modified_gmt"}
+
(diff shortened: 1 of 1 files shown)


borekb:
👋 @abtris, thanks for the PR! @JanVoracek will be looking at this later this week, quick notes from yesterday:

- The driver will probably be best managed as a separate project in this repo, outside of `plugins/versionpress/src`. There'll need to be a dev setup, compilation instructions, updated build script, etc.
- Isolated tests would be tricky, we might be happy enough exercising the driver as part of our full test suite.
- Would CI be building the driver?
- All files are loaded in memory by the current implementation, this should be converted to streaming.

abtris:
I agree separate repo is good idea. Github Action or CI can build with [goreleaser](https://github.com/goreleaser/goreleaser). Before we change implementation I will go with benchmark.

Issue #1259, opened 2017-07-04
Travis CI


borekb:
With Dockerized tests (#1227) we're ready to set up Travis CI for this repo. Here's their docs page: [Using Docker in Builds - Travis CI](https://docs.travis-ci.com/user/docker/).

---

🏁 

- [x] Initial implementation was #1381. We reverted it in #1394 as the tests were failing at that point.
- [x] Tests were improved in #1389 and are passing since #1400.
- [ ] With that, there's a new PR adding Travis CI build, #1415.

borekb:
Resolved by #1381. 🎉

borekb:
I've reverted #1381 with #1394 for reason given in that PR:

> I'm going to revert the Travis CI setup temporarily as all the builds fail right now (#1383) and it's not very useful to have CI running at this point. Quite to the contrary, e.g., Markdown changes that are fine are flagged as failed even though they're correct.

I'm also going to reopen this issue and based on the experience that #1381 gave us, update the issue description.

borekb:
Travis also [added Windows support](https://blog.travis-ci.com/2018-10-11-windows-early-release) recently, that would be useful to set up as well.

borekb:
Moving the following notes from the issue description here as we won't implement it that properly for now:

---

**UPDATE 03/2019**: Based on our experience with #1381, it's now clearer what we want from a CI setup:

- The builds should respect that this repo is a monorepo and only check what needs to be checked. For example, with #1381 setup, if I updated a Markdown file in the `docs` folder, full tests were still started (they used to take over an hour at that point), Docker images were built (incl. pulling all the base layers), etc. which is both wasteful and also makes it harder to debug the actual issue as the reported problems might come from different parts of the repo.

    Examples of separated builds:
    
    - Changes in `docs` trigger Markdown linting & Netlify preview deployment.
    - Changes in `plugins/versionpress` run full tests.
    - Changes in `frontend` run `tsc`, TSLint and frontend tests.
    - Changes in `dev-env/wordpress-[cli-]image` trigger Docker build.

    Multiple paths can be affected in a single PR which should trigger all related CI tasks, for example, if a UI is updated and there's a new screenshot in the docs, points 1 and 3 from above should run.

[...]

PR #1394, opened 2019-03-09
Revert "Run tests in Travis" for now

borekb:
Reverts versionpress/versionpress#1381

I'm going to revert the Travis CI setup temporarily as all the builds fail right now (#1383) and it's not very useful to have CI running at this point. Quite to the contrary, e.g., Markdown changes that are fine are flagged as failed even though they're correct.

I'm going to describe how I'd imagine a good CI setup for us in a separate issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,37 +2,18 @@ sudo: false
 language: php
 php:
   - 7.2
-services:
-  - docker
-addons:
-  apt:
-    packages:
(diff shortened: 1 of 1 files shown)


borekb:
Learning from this added to https://github.com/versionpress/versionpress/issues/1259#issue-240339268. It was very useful!

Issue #1389, opened 2019-03-07
Improve Dockerized testing infrastructure


borekb:
Full tests (`npm run tests:full`) currently take over an hour to run on a relatively powerful machine (Core i7 MacBook Pro) and there are some other issues related to tests as we reviewed them with @JanVoracek today.

**UPDATE**: In the end, this issue is a relatively major update of the dockerized dev / testing environment last updated in #1329.

### Highlights

- Full tests now run _much_ faster thanks to volumes instead of host-mounted folders. On my machine, the time dropped from **1 hour 15 minutes** to just **11 minutes**. #1396
- Permissions between containers (Debian vs. Alpine) have been fixed. #1397
- New `run-tests.ts` script that orchestrates how tests run. For example, it decides when to start a WordPress stack (if at all), cleans up after itself, etc. #1401
- The main testing scripts is now just `npm run tests` (the `:full` suffix has been dropped).
    - Also, all the `tests:custom` scripts have been dropped as `run-tests.ts` now accepts parameters, for example:<br>`npm run tests -- --testsuite Unit --filter CursorTest`
- Docker images are now fixed using their digests. No more surprises if upstream images change. #1398
- Some effort has been put into testing and documenting how test work on Docker Toolbox. #1403

During the work, I've also encountered several other things that are noted in the "Deferred items" section below.

### TODO list

- [x] Split testing documentation from Dev Setup, it's quite long on its own. #1392
    - [x] ... and smaller related content changes. #1393
- [x] Split docker-compose services into two files, `docker-compose.yml` and `docker-compose.tests.yml`. #1395
- [x] ⭐ Permissions: Run CLI image as UID 33 (Debian's "www-data") #1397
- [x] ⭐ Volumes: use named volumes (i.e., don't map to host OS) & have some way to inspect the data later. #1396
- [x] Use Docker image digests instead of tags. #1398
    - They are uglier, e.g., `busybox:latest` becomes `busybox@sha256:38a203e1986cf79639cfb9b2e1d6e773de84002feea2d4eb006b52004ee8502d`, but it's the only way to make runs predictable (docker-compose lacks a concept of a lock file).
    - This has already been considered in https://github.com/versionpress/versionpress/pull/1329 but we eventually didn't go for it.
    - `docker-compose config --resolve-image-digests` can help with that.
    - Awesome post about the problem and solutions: [Overcoming Docker’s mutable image tags](https://renovatebot.com/blog/docker-mutable-tags)
[...]

PR #1396, opened 2019-03-09
Use named volumes for tests

borekb:
This is one of the major items of #1389 that should lead to much faster test runs and to a better encapsulation of tests (fewer host mounts). It has already been drafted by @candrews in https://github.com/versionpress/versionpress/pull/1380, this PR delivers it with all the details.

- [x] Use a named volume for `/var/www/html`
- [x] Named volume for logs – I'm currently hitting the same error that @candrews reported in #1379, let's see...
- [x] Document ways to inspect WordPress files and site logs after tests finish.
- [x] Unify volume names (snake_case vs. dashed-names)

Useful resource: [Manage data in Docker](https://docs.docker.com/storage/) from Docker Documentation – discusses differences between volumes and "bind mounts" (host OS mounts).

### DIFF
--- a/dev-env/wordpress-cli-image/Dockerfile
+++ b/dev-env/wordpress-cli-image/Dockerfile
@@ -1,5 +1,11 @@
 FROM wordpress:cli
 
+# IMPORTANT: this image is Apline-based where www-data is UID 82 while in Debian-based WordPress image,
+# it's 33. Both will be reading and writing to `/var/www/html` so we'll be using UID 33 in this Dockerfile
+# to avoid any permission issues. See e.g. docker-library/wordpress#256.
+#
+# Never use www-data in this Dockerfile!
+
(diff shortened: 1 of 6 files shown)


borekb:
https://github.com/versionpress/versionpress/pull/1396/commits/718a29f08dbce983b2955153ac063498cdf6f1c2 implements a named volume for `/var/www/html` and the speedup is really significant:

```
npm run tests:stop-and-cleanup
npm run tests:custom-with-wordpress -- -c phpunit.xml --filter End2End
```

Before: 41m 33s
After 718a29f08d: **7m 45s**

🎉

borekb:
5369b0caa8 merges in the important PR #1397 that fixed www-data permissions.

borekb:
PR description updated, I'm going to merge this.

PR #1397, opened 2019-03-10
Run CLI image as UID 33 (Debian's "www-data")

borekb:
The CLI image is based on Alpine where the www-data user is UID 82 while the main WordPress image is based on Debian where www-data stands for UID 33. This causes trouble as files to `/var/www/html` are attempted to being read and written by different UIDs, depending on whether it's the Apache web server or WP-CLI / our testing code.

PR https://github.com/versionpress/versionpress/pull/1388 fixed the user on the `docker-compose.yml` level but that caused [this portion of Dockerfile](https://github.com/versionpress/versionpress/blob/618813bf79d9c2dc63bf5ebef751c84c46323130/dev-env/wordpress-cli-image/Dockerfile#L21-L32) to become invalid. This PR moves the solution to the Dockerfile level – it explicitly uses "33" instead of "www-data".

Some related resources:

- https://github.com/docker-library/wordpress/issues/256
- https://github.com/docker-library/wordpress/issues/279
- https://github.com/docker-library/docs/pull/1284/commits/10ba46a559b0f8853660739ea03ed1f99d144226

(Work on this is part of #1389.)

### DIFF
--- a/dev-env/wordpress-cli-image/Dockerfile
+++ b/dev-env/wordpress-cli-image/Dockerfile
@@ -1,12 +1,8 @@
 FROM wordpress:cli
 
-# IMPORTANT: About users
-#
-# This image is Alpine-based but needs to interact with Debian-based WordPress image. The catch is
-# that "www-data" is a different UID on the two systems – 33 in Debian but 82 on Alpine Linux.
-#
-# To make everything work smoothly together, we explicitly use UID 33 here (which could also
-# be written as the 'xfs' user).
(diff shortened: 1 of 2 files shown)


borekb:
I've ran `npm run build-images` and then `npm run tests:full`, it seems to work equally well as #1388 but also fixes the issues with writing to test-logs that I've encounter in #1396 (similar to #1379).

@candrews, could you please review this change?

borekb:
It's working for me and it'll be useful to have this in `master` for other PRs around #1389 so I'm going to merge this without further verification. We can do a followup PR if needs be.

@candrews, I'd still be interested in your opinion on this.

PR #1414, opened 2019-03-26
Fix markdownlint violation in docs/content/en/troubleshooting/iis.md

candrews:
@borekb simple change so markdownlint passes

PR #1415, opened 2019-03-26
Have Travis run tests

candrews:
Issue: #1259

Travis will now run markdownlint, php codesniffer, build the docker images, and run the full tests.

With this change, VersionPress can now only merge PRs that have passing tests, ensuring that all tests always pass.

(This is a similar PR to #1381, after we have fixed the testing infrastructure in #1389 so that the tests actually pass.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,6 @@ before_install:
 
 install:
   - npm ci
-  - npm run init-dev
 
 before_script:
   - sudo /etc/init.d/mysql stop # Travis runs mysql by default and our docker tests want the mysql port (3306), so stop mysql
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,18 +2,36 @@ sudo: false
 language: php
 php:
   - 7.2
-
-branches:
-  only: 
-    - master
-
+services:
+  - docker
+addons:
+  apt:
+    packages:
+      - docker-ce
 cache:
   directories:
     - $HOME/.composer/cache
-
+    - $HOME/docker
+    - $HOME/.npm
 before_install:
-  - git config --global github.accesstoken $GITHUB_OAUTH_TOKEN
-  - composer config -g github-oauth.github.com $GITHUB_OAUTH_TOKEN --no-interaction
+  - sudo rm /usr/local/bin/docker-compose
+  - curl -L https://github.com/docker/compose/releases/download/1.23.2/docker-compose-`uname -s`-`uname -m` > docker-compose
+  - chmod +x docker-compose
+  - sudo mv docker-compose /usr/local/bin
+  - nvm install 8
+  - nvm use 8
+
+install:
+  - npm ci
+  - npm run init-dev
+
+before_script:
+  - sudo /etc/init.d/mysql stop # Travis runs mysql by default and our docker tests want the mysql port (3306), so stop mysql
 
 script:
-    cd plugins/versionpress && composer install && ./vendor/bin/phpcs --standard=ruleset.xml
+  - set -e
+  - npm run lint:markdown
+  - npm run lint:php
+  - npm run build-images
+  - npm run tests
+
--- a/package.json
+++ b/package.json
@@ -24,7 +24,8 @@
     "_push-image-cli": "docker push versionpress/wordpress:cli",
     "get-image-digests": "node -r ./scripts/node_modules/ts-node/register scripts/get-dockerhub-digests.ts versionpress/wordpress",
     "postinstall": "(cd scripts && npm install) && npm run init-dev",
-    "lint:markdown": "markdownlint -i '{**/vendor/**,**/node_modules/**,ext-libs/**,dev-env/**}' ."
+    "lint:markdown": "markdownlint -i '{**/vendor/**,**/node_modules/**,ext-libs/**,dev-env/**}' .",
+    "lint:php": "./plugins/versionpress/vendor/bin/phpcs --standard=./plugins/versionpress/ruleset.xml"
   },
   "engines": {
     "node": ">= 8.11",


candrews:
I included the commit from https://github.com/versionpress/versionpress/pull/1414 so Travis actually passes :)

borekb:
(Thanks for rebasing 😄.)

borekb:
Have you seen the updated description o #1259? While the tests run much faster now, it's still wasteful to run long builds for commits that didn't change any code.

candrews:
IMHO, Travis currently takes 15 minutes to run the whole thing, which isn't that long. It's not an ideal allocation of compute resources, but I don't see the harm - a few extra occurrences of 15 minute builds isn't going to really impact global warming, and since the builders are free, it's not going to break the bank either :)

Plus, if you run full tests for each PR, you always know the full tests pass. If you get "fancy" and try to run a subset, it's possible to make a mistake and not run a test that should have been run due an error in the test harness. I think KISS is the way to go.

In this PR, the tests also fast fail (that's due to the `set -e` in the `scripts` section) so if there's a failure, it'll be reported as quickly as possible and the rest of the steps don't run. I also ordered it so the fastest stuff is first.

borekb:
Ok, let's do it for now. If we want to improve this in the future, here are some links:

- [Simple update of `.travis.yml`](https://twitter.com/keeskluskens/status/996005160089477120). [This](https://github.com/alisw/ali-bot/blob/cd9bd1ec3f88c3d267f74e10a9addfd710054ba7/.travis.yml) is similar and [this is how Kubernetes does it](https://github.com/kubernetes-sigs/federation-v2/pull/584/commits/c36c0a2bf1a639dc241ec2ce010f9dc9dc3dcdd5).
- [This Neovim config example](https://github.com/epeli/neovim-config/blob/94c5039847780610eed0da66b9add08ca2241a11/plugged/neomake/.travis.yml) looks a bit more complex but I _think_ they're achieving quite an excellent setup.
- I'd prefer to have it as a shell script so that I can run all the validations locally. Here are some examples of how to do that:
    - [DZone article](https://dzone.com/articles/build-test-and-deploy-apps-independently-from-a-mo) (the key being <code>folders=&#96;git diff --name-only $SHIPPABLE_COMMIT_RANGE | sort -u | awk 'BEGIN {FS="/"} {print $1}' | uniq&#96;</code> )
    - Example [1](https://gist.github.com/essmahr/05f37b0dec779c5d01b9225241e7d208), [2](https://github.com/forforeach/attracker/blob/33b1669933885b15c4f953cd0327ae4a4dbe1754/scripts/test.sh), [3](https://github.com/chadgrossman/qmk_firmware/blob/00596d55e34c003d55a530366bedfb2ea467bedc/util/travis_build.sh), [4](https://github.com/faucetsdn/faucet/blob/dfa86d2836cc73ed46273ebc6fe30f9784a5f0a0/travis/runtests.sh).

This is the upstream issue: https://github.com/travis-ci/travis-ci/issues/6301.

One thing I currently don't like is that Docker images are re-built every time – looks like some layer caching is missing or something (I don't really know how the Travis VM caches things but for example `npm install` or `composer install` run quickly on subsequent runs). This is not only inefficient but also produces a lot of noise in Travis logs. Could this be improved?

borekb:
Overall, it's awesome to see the test passing on a CI! 🤖

.travis.yml
```diff
@@ -2,18 +2,39 @@ sudo: false
...
-  only: 
-    - master
+services:
+  - docker
+addons:
+  apt:
+    packages:
+      - docker-ce
+env:
+  - DOCKER_COMPOSE_VERSION=1.23.2
```
borekb: I'd just hardcode this, similarly to the Node version.

.travis.yml
```diff
@@ -2,18 +2,39 @@ sudo: false
...
+  - sudo rm /usr/local/bin/docker-compose
+  - curl -L https://github.com/docker/compose/releases/download/${DOCKER_COMPOSE_VERSION}/docker-compose-`uname -s`-`uname -m` > docker-compose
+  - chmod +x docker-compose
+  - sudo mv docker-compose /usr/local/bin
+  - nvm install 8
+  - nvm use 8
+
+install:
+  - npm ci
+  - ( cd plugins/versionpress && composer install )
```
borekb: `composer install` runs as part of init-dev, doesn't it?

.travis.yml
```diff
@@ -2,18 +2,39 @@ sudo: false
...
+
+before_script:
+  - sudo /etc/init.d/mysql stop # Travis runs mysql by default and our docker tests want the mysql port (3306), so stop mysql
 
 script:
-    cd plugins/versionpress && composer install && ./vendor/bin/phpcs --standard=ruleset.xml
+  - set -e
+  - npm run lint:markdown
+  - ( cd plugins/versionpress && ./vendor/bin/phpcs --standard=ruleset.xml )
```
borekb: This could be a `package.json` script as well, to be consistent with the rest. I'd call it `phpcs` or `lint:php`.

.travis.yml
```diff
@@ -2,18 +2,39 @@ sudo: false
...
+  apt:
+    packages:
+      - docker-ce
+env:
+  - DOCKER_COMPOSE_VERSION=1.23.2
 
 cache:
   directories:
     - $HOME/.composer/cache
+    - "$HOME/.npm"
```
borekb: Docker layer caching should be added as well if possible, see https://github.com/versionpress/versionpress/pull/1415#issuecomment-476875059.

.travis.yml
candrews: done!

.travis.yml
candrews: Changed to run init-dev in the install phase instead of using composer install

.travis.yml
candrews: done!

.travis.yml
candrews: I implemented this approach: https://github.com/travis-ci/travis-ci/issues/5358#issuecomment-248915326

borekb:
Thanks for the changes, I'll take a look shortly.

I'd just like to ask for "normal commits" in the future, not force-pushes. For example, the `.travis.yml` evolved over several commits which is not evident from the history of this PR, and some of the links are broken:

Screenshot 2019-03-27 at 15 49 06

Also, the `iis.md` formatting fix was first here, then it wasn't (the first force-push), now it's back again (maybe a mistake during the second force-push?), it's all quite confusing.

We prefer multiple smaller commits, some of them heavy WIP, over "clean" (but fake) history.

.travis.yml
borekb: `init-dev` runs as a post-install lifecycle script, it's not necessary to call it manually. See [Travis logs](https://travis-ci.org/versionpress/versionpress/builds/511671083#L531-L648).

.travis.yml
borekb: 👍

candrews:
> We prefer multiple smaller commits, some of them heavy WIP, over "clean" (but fake) history.

No problem - different projects have different requirements, and I've recently been working on ones that use a different model :)

.travis.yml
borekb: The latest build is still slow but that might be because the cache was only being created? Let's see in the newer builds (I saw you just triggered one...).

Screenshot 2019-03-27 at 16 03 30

borekb:
Thank you, it's not documented very well right now, our fault ;)

.travis.yml
borekb: The build is now failing, marking is as not resolved.

candrews: How about we merge this now (commit bb928) and work on the docker caching in a separate PR as an improvement?

borekb: Agree, I'd also do it that way. Can you please update this PR so that the build passes?

candrews: Done!

borekb: Thanks. Just remember, please no force-pushing 😄.

borekb:
Thanks, @candrews! I've created a separate issue for Docker caching: #1419.

Issue #1420, opened 2019-04-10
Tracking original IDs or general order


pavelevap:
During work on test infrastructure we found a problem with order of posts on cloned site. When there are several posts (or other items) **published at the same time and ordered by that time**, their **order can be different on original and cloned site**, see example in https://github.com/versionpress/versionpress/issues/1409#issuecomment-480819108. 

Order on original site is based on ID (order from database storage), but cloned site depends on loading `.ini` files through iterators:

https://github.com/versionpress/versionpress/blob/e2209836d30889727991365fde2c983d0121e5a1/plugins/versionpress/src/Storages/DirectoryStorage.php#L157-L168

And VersionPress ignores original IDs, so there is currently no way to order posts by their original ID on cloned sites.

Maybe there is a time to start **tracking original IDs** (VP could make a clean clone with original IDs which could be helpful in some cases) or **general order** (VP could generally know order of created items at least).

borekb:
We spent some time chatting about this with @JanVoracek, nothing too deep or well-explored yet but to record the key points:

- Storing auto-increment IDs is probably not doable. Any re-numbering of IDs would lead to changes in the `vpdb` files, leading to new commits, leading to strange Git history at the very least.
- Order represented by precise time of creation is probably doable. (I'll be calling it 'timestamps' below but no specific implementation is assumed yet.)
- There are generally two places to store the timestamp: file name (path) or file contents.
    - Path storage should be faster to work with as there would be no need to read contents of files.
        - Prefixing VPID with timestamp or incorporating it into the beginning of it would have a problem that files would not be evenly distributed to 256 folders. (VersionPress uses first two letters of VPID as a subfolder under `vpdb/<entity>` to deal with situations where thousands or even millions of files would end up in a single folder.) So it would still be some challenge to order by timestamp correctly but generally, it should be more efficient than reading files contents.
    - Storing timestemp in file contents feels a bit more flexible, maybe we'd even need it for things like postmeta. But also possibly inefficient.
        - Performance is an important concern as, for example, restoring site from Git or creating a staging environment will need to take all timestamps into account. In the worst case, this means parsing a lot of files and building a large data structure in memory.
- We acknowledge this is an issue that should be solved one day but don't see it as a high priority. There haven't been reports about this yet in the support repo or other channels.

PR #1421, opened 2019-04-10
Fix cloneLooksExactlySameAsOriginal test

### LINKED ISSUE #1409: `CloneMergeTest::cloneLooksExactlySameAsOriginal` fails
After improving the test runs reliability in #1389 / #1401, one test is still failing consistently: [`CloneMergeTest::cloneLooksExactlySameAsOriginal`](https://github.com/versionpress/versionpress/blob/1a21fcb11b2693b8eac44b95afd904cfde13e517/plugins/versionpress/tests/Workflow/CloneMergeTest.php#L34-L49).

When Workflow tests run separately, or before End2End tests (see https://github.com/versionpress/versionpress/issues/1400#issuecomment-4719877160), this test passes but on a site exercised by End2End tests, it fails. Upon initial exploration, @pavelevap found out that indeed the clone looks different than the original.

I'm going to implement a workaround to make the tests pass but we should return to this and fix the underlying problem.

- [ ] Update CloneMergeTest to run from a clean WP site (it's better than skipping the test entirely). @borekb
- [ ] Find out what the problem is and implement a fix. @pavelevap

pavelevap:
Resolves #1409 

The cause of this problem depends on #1420, this PR will be only workaround to avoid posts with different titles published at the same time.

### DIFF
--- a/plugins/versionpress/tests/Automation/WpAutomation.php
+++ b/plugins/versionpress/tests/Automation/WpAutomation.php
@@ -135,11 +135,15 @@ class WpAutomation
      *
      * @see wp_insert_post()
      * @param array $post (as wp_insert_post)
+     * @param bool $waitOneSecond Wait to ensure that there are no testing posts published at the same time.
+     *                            Currently used only for End2End tests which would fail without it, see #1420.    
      * @return int
      */
-    public function createPost(array $post)
+    public function createPost(array $post, $waitOneSecond = false)
     {
-        sleep(1); // Ensure that there are no testing posts published at the same time, see issue #1420
+        if ($waitOneSecond) {
+            sleep(1);
+        }
         $post["porcelain"] = null; // wp-Cli returns only id
         return intval($this->runWpCliCommand('post', 'create', $post));
     }
--- a/plugins/versionpress/tests/Automation/WpAutomation.php
+++ b/plugins/versionpress/tests/Automation/WpAutomation.php
@@ -139,6 +139,7 @@ class WpAutomation
      */
     public function createPost(array $post)
     {
+        sleep(1); // Ensure that there are no testing posts published at the same time, see issue #1420
         $post["porcelain"] = null; // wp-Cli returns only id
         return intval($this->runWpCliCommand('post', 'create', $post));
     }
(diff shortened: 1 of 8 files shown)


pavelevap:
After partial revert of PR #1410, tests are failing again:

https://travis-ci.org/versionpress/versionpress/builds/518248095#L1653

image

Date for "Test post for comments" has to be changed to preserve order on cloned site.

borekb:
There was an issue with the exit code from the `run-tests.ts` script:

Screenshot 2019-04-10 at 14 30 52

This is now resolved by #1422.

pavelevap:
OK, this PR is correctly broken now and tests are failing :-)

pavelevap:
This is as small as possible fix to make all tests pass again, ready for discussion and review.

I started with changing `post_date` for some posts, but it would lead to many manual changes to enforce the right order. So I ended with removing predefined `post_date` in End2End tests and ensuring that there are no posts published at the same time.

plugins/versionpress/tests/Automation/WpAutomation.php
```diff
@@ -135,11 +135,15 @@ public function uninstallVersionPress()
      *
      * @see wp_insert_post()
      * @param array $post (as wp_insert_post)
+     * @param bool $waitOneSecond Wait to ensure that there are no testing posts published at the same time.
+     *                            Currently used only for End2End tests which would fail without it, see #1420.    
      * @return int
      */
-    public function createPost(array $post)
+    public function createPost(array $post, $waitOneSecond = false)
```
borekb: Though it's a step in the right direction, this is still possible:

```php
$post1["post_date"] = "2011-11-11 11:11:11";
$post2["post_date"] = "2011-11-11 11:11:11";

$wpAutomation->createPost($post1, true);
$wpAutomation->createPost($post2, true);

// date collision not really prevented
```  

I think the method could look something like this:

```php
public function createPost(array $post, $managePostDateAutomatically = false)
{
    if ($managePostDateAutomatically) {
        if (isset($post["post_date"])) {
            throw new Exception('post_date must not be set when managing post date automatically');
        }
        sleep(1);
    }
    // ...
}
```

The usage of `sleep` becomes an implementation detail that could change in the future, the main thing is that the method guarantees certain behavior.

borekb:
We had a discussion about this with @pavelevap and @JanVoracek yesterday and decided that the approach from #1410 is actually preferable as it runs workflow tests from a well-known state.

It's true that running the clone & merge tests after End2End tests allowed us to uncover an important detail about VersionPress' behavior, documented in #1420, however, we shouldn't be using a combination of End2End and Workflow tests to verify that.

So I'm going to close this PR and take a look at a change done in #1410, it will probably need some minor updates in comments but overall, that should be our solution for #1409. @pavelevap, thanks for the effort anyway, it lead us to think more deeply about how tests should be written.

PR #1429, opened 2019-04-23
Capitalized values aren't handled correctly

candrews:
VersionPress doesn't correctly handled capitalized values in its schema configuration. The intent appears to be that comparison should be case insensitive, but that's not how it currently works.

For example, this configuration:
```
some-entity:
    ignored-entities:
        capitalized_value_field: VALUE
```

does not ignore this entity:
```
some-entity:
    capitalized_value_field: VALUE
```

and it should.

### DIFF
--- a/plugins/versionpress/src/Utils/QueryLanguageUtils.php
+++ b/plugins/versionpress/src/Utils/QueryLanguageUtils.php
@@ -83,7 +83,7 @@ class QueryLanguageUtils
 
                 if ($isWildcard && preg_match(QueryLanguageUtils::tokensToRegex($valueTokens), $entity[$field])) {
                     return true;
-                } elseif (strval($entity[$field]) === $value) {
+                } elseif (strtolower(strval($entity[$field])) === strtolower($value)) {
                     return true;
                 }
 
--- a/plugins/versionpress/src/Utils/QueryLanguageUtils.php
+++ b/plugins/versionpress/src/Utils/QueryLanguageUtils.php
@@ -83,7 +83,7 @@ class QueryLanguageUtils
 
                 if ($isWildcard && preg_match(QueryLanguageUtils::tokensToRegex($valueTokens), $entity[$field])) {
                     return true;
-                } elseif (strtolower(strval($entity[$field])) === strtolower($value)) {
+                } elseif (strtolower(strval($entity[$field])) === $value) {
                     return true;
                 }
 
--- a/plugins/versionpress/tests/Unit/EntityInfoTest.php
+++ b/plugins/versionpress/tests/Unit/EntityInfoTest.php
@@ -147,7 +147,6 @@ class EntityInfoTest extends \PHPUnit_Framework_TestCase
         $entity = [
             'some_field' => 'value',
             'other_field' => 'b',
-            'capitalized_value_field' => 'qqq',
         ];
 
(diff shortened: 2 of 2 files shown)


pavelevap:
Hi, @candrews,

thank you, I can confirm this issue. I tried to add following to `ignored-entities` for `postmeta` into `schema.yml`:

```
    - 'meta_key: TEST'
```

Then I tried to add this postmeta in WP administration and `TEST` is tracked even if it should not.

pavelevap:
Problem was caused by this commit: https://github.com/versionpress/versionpress/commit/2330858026caa39843b714eb7a2507884fcecc4e.

See also following comment: https://github.com/versionpress/versionpress/pull/730#issuecomment-200032214

Values in `QueryLanguageUtils::createRulesFromQueries` were lowercased which helped filtering commits in WP administration, but broke case sensitive values.

plugins/versionpress/src/Utils/QueryLanguageUtils.php
```diff
@@ -83,7 +83,7 @@ public static function entityMatchesSomeRule($entity, $rules)
 
                 if ($isWildcard && preg_match(QueryLanguageUtils::tokensToRegex($valueTokens), $entity[$field])) {
                     return true;
-                } elseif (strval($entity[$field]) === $value) {
+                } elseif (strtolower(strval($entity[$field])) === strtolower($value)) {
```
pavelevap: `$value` will be lowercased automatically because of this: https://github.com/versionpress/versionpress/blob/4475b80e305d3ca46bff7b3770a5e36ef56fc6a5/plugins/versionpress/src/Utils/QueryLanguageUtils.php#L47-L49

candrews: Thanks! I updated this PR to remove the unnecessary `strtolower` that I added.

candrews:
@borekb can this now be merged?

borekb:
Hi @candrews, yes, thank you for the fix and updated tests! The only thing I'm not sure about is "qqq" in the tests but I'll leave that as a review comment.

BTW, this sparked a relatively deep debate between @pavelevap, myself and @JanVoracek about how we implement case insensitivity in search and frequently written / ignored entities. @pavelevap will be creating a separate issue for that, the fix in this PR is fine and can be merged soon.

plugins/versionpress/tests/Unit/EntityInfoTest.php
```diff
@@ -129,6 +147,7 @@ public function commonEntityIsNotFalselyIdentifiedAsIgnored()
         $entity = [
             'some_field' => 'value',
             'other_field' => 'b',
+            'capitalized_value_field' => 'qqq',
```
borekb: This test method (overall) seems to verify that more complex rules with multiple conditions, e.g., `'some_field:"value" other_field:"a"'`, match only when all their parts are satisfied. ("Common entity" in the test name is not helpful to me but I struggle coming up with a better suggestion 😄.)

The capitalized field example is a separate rule, `'capitalized_value_field: VALUE'`, so I don't think it belongs here.

candrews: Makes sense to me - I've removed that line from the test.

borekb:
Thanks for this, @candrews!

PR #1437, opened 2019-05-03
 Update phpmyadmin/sql-parser to 4.3.1

candrews:
VersionPress was using 3.4.0; the latest version is 4.3.1.

Based on the changelog, this seems like a good idea: https://github.com/phpmyadmin/sql-parser/blob/v4.3.1/CHANGELOG.md

### DIFF
--- a/plugins/versionpress/composer.json
+++ b/plugins/versionpress/composer.json
@@ -22,7 +22,7 @@
     "symfony/process": "^3.3",
     "michelf/php-markdown": "^1.7",
     "symfony/filesystem": "^3.3",
-    "phpmyadmin/sql-parser": "3.4",
+    "phpmyadmin/sql-parser": "^4",
     "tracy/tracy": "^2.4"
   },
   "require-dev": {
--- a/plugins/versionpress/composer.json
+++ b/plugins/versionpress/composer.json
@@ -29,7 +29,7 @@
     "phpunit/phpunit": "^5.7",
     "squizlabs/php_codesniffer": "^3.0",
     "fzaninotto/faker": "^1.6",
-    "mikey179/vfsStream": "^1.6",
+    "mikey179/vfsstream": "^1.6",
     "10up/wp_mock": "^0.1.1",
     "phpunit/phpunit-selenium": "^3.0",
     "icecave/semver": "^3.0"
--- a/plugins/versionpress/composer.json
+++ b/plugins/versionpress/composer.json
@@ -22,7 +22,7 @@
     "symfony/process": "^3.3",
     "michelf/php-markdown": "^1.7",
     "symfony/filesystem": "^3.3",
-    "phpmyadmin/sql-parser": "^4",
+    "phpmyadmin/sql-parser": "^4.3",
     "tracy/tracy": "^2.4"
   },
   "require-dev": {
--- a/plugins/versionpress/composer.lock
+++ b/plugins/versionpress/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "272518869b7fc998d665619af5ea7cf4",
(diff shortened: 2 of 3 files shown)


candrews:
It appears some tests need updating (or some problems need fixing).

pavelevap:
I agree, we had similar problem recently with older `phpmyadmin/sql-parser` version, see https://github.com/versionpress/support/issues/164#issuecomment-486385745

plugins/versionpress/composer.json
```diff
@@ -22,7 +22,7 @@
     "symfony/process": "^3.3",
     "michelf/php-markdown": "^1.7",
     "symfony/filesystem": "^3.3",
-    "phpmyadmin/sql-parser": "3.4",
+    "phpmyadmin/sql-parser": "3.4.17",
```
pavelevap: Maybe we should use caret sign here `^3.4` (as for other dependencies)?

pavelevap:
Hmm, this will be probably tricky. Our tests are broken from `phpmyadmin/sql-parser` 3.4.9 when support for `DELETE` queries was added, see https://github.com/phpmyadmin/sql-parser/pull/87

But our tests are using `DELETE ... JOIN` queries, for example here:

https://github.com/versionpress/versionpress/blob/380f41d55d106d9f45bd0c4931cbac7eb0e3dac3/plugins/versionpress/tests/Unit/SqlQueryParserTest.php#L131-L132

And it was supported/fixed later in version 4.1.3, see https://github.com/phpmyadmin/sql-parser/issues/144

So, safe version for us is 3.4.8 or we could switch higher (at least 4.1.3) which would need some more work, I guess (probably adjust for namespace handling in 4.0.0).

candrews:
This PR now updates sql-parser to the latest version, 4.3.1. All tests pass :+1:

borekb:
Nice. @pavelevap, if you're happy with the change, feel free to merge it (I won't do it myself because maybe you'll want to verify this change on a real-world WP installation; bumping a major / breaking version of an important dependency is always a bit scary 😄).

candrews:
@borekb @pavelevap _bump_

borekb:
I assume @pavelevap will take a look shortly, I'm personally happy with this PR.

pavelevap:
I like it too :+1:  I will try it tomorrow on my testing site and then merge.

pavelevap:
I tried `npm i` and I can still see an error related to `mikey179/vfsStream` even if there is a commit to fix it: https://github.com/versionpress/versionpress/pull/1437/commits/5cf51fabcb8c8cf0b19fd843e8f94f57065d54f0.

`Deprecation warning: require-dev.mikey179/vfsStream is invalid, it should not contain uppercase characters. Please use mikey179/vfsstream instead. Make sure you fix this as Composer 2.0 will error.`

plugins/versionpress/composer.json
```diff
@@ -22,7 +22,7 @@
     "symfony/process": "^3.3",
     "michelf/php-markdown": "^1.7",
     "symfony/filesystem": "^3.3",
-    "phpmyadmin/sql-parser": "3.4",
+    "phpmyadmin/sql-parser": "^4",
```
pavelevap: Should we better use common entry `^4.3` instead of only `^4` (as all other dependencies in this file)?

candrews: sql-parser follows semver so all future 4.x versions will work with VersionPress. I don't see any value in restricting the minor version, especially since VP uses a lock file so all dependencies are fixed and any dependency updates must be reviewed / are not automatic. However, if that's what you want to do, I'm happy to make that change to get this PR merged. :)

borekb: `^4.3` doesn't really restrict the minor version (as you pointed out), it's just a small stylistic request to keep things consistent.

The bigger issue is the `npm install` warning, there are some hints here: https://github.com/bovigo/vfsStream/pull/174.

candrews: I added a commit to fix that warning.

pavelevap:
Thank you, @candrews!

PR #1438, opened 2019-05-03
Use stripcslashes instead of stripslashes when handling values in sql

### LINKED ISSUE #1436: Slashes are stripped from values in INSERT queries, breaking some data
To reproduce this issue, create a new file called `test.php`:
```
<?php
require( dirname( __FILE__ ) . '/wp-load.php' );
update_option( 'lines6', 'First row
Second row
Third row' );
```

Then run it by hitting the URL in the browser.

Actual:
```
$ cat wp-content/vpdb/options/li/lines6.ini 
[lines6]
option_value = "First rownSecond rowXnThird row"
autoload = "yes"
```
Expected:
```
$ cat wp-content/vpdb/options/li/lines6.ini 
[lines6]
option_value = "First row
Second row
Third row"
autoload = "yes"
```

Interesting, executing `test.php` a second time results in the expected data being written to the ini file. So the issue is only happening for INSERTs (not for UPDATEs).

candrews:
Resolves #1436

For example, in query:
```
INSERT INTO `wp_options` (`option_name`, `option_value`, `autoload`) VALUES ('lines', 'First row\nSecond row\nThird row', 'yes')
```
the extracted value should be:
```
First row
Second row
Third row
```
But, before this change, due to the use of `stripcslashes`, the value would be:
```
First rownSecond rownThird row
```

### DIFF
--- a/plugins/versionpress/src/Database/SqlQueryParser.php
+++ b/plugins/versionpress/src/Database/SqlQueryParser.php
@@ -204,7 +204,7 @@ class SqlQueryParser
         if ($sqlStatement instanceof UpdateStatement) {
             $dataSet = [];
             foreach ($sqlStatement->set as $set) {
-                $dataSet[str_replace('`', '', $set->column)] = stripslashes($set->value);
+                $dataSet[str_replace('`', '', $set->column)] = stripcslashes($set->value);
             };
             return $dataSet;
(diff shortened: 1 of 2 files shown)


borekb:
Nice, thank you!

PR #1444, opened 2019-05-12
Reorganize case sensitivity for matching entities and filtering commits

### LINKED ISSUE #1439: Rethink case sensitivity comparison for entities
History:
- Values for `ignored-entities` and `frequently-written` were originally compared as **case-sensitive**.
- But during work on issue #586 were values changed to **case-insensitive** to handle commit filtering in `QueryLanguageUtils::createRulesFromQueries`, see following commits: https://github.com/versionpress/versionpress/commit/2330858026caa39843b714eb7a2507884fcecc4e and https://github.com/versionpress/versionpress/commit/b54109f51fd2945e606ba8ddb1e6f6e2450893f1.
- Recognizing capital values for entities was fixed in #1429.
- And now we should rethink usage of `strtolower()` in `QueryLanguageUtils::createRulesFromQueries` and `QueryLanguageUtils::entityMatchesSomeRule`.

Plan:
- [ ] Move case-insensitive comparison from `QueryLanguageUtils::createRulesFromQueries` to `QueryLanguageUtils::entityMatchesSomeRule`.
- [ ] Replace `strtolower()` with `mb_strtolower()` - values can be defined in different languages.
- [ ] Update documentation: https://github.com/versionpress/versionpress/blob/4475b80e305d3ca46bff7b3770a5e36ef56fc6a5/docs/content/en/developer/plugin-support.md#frequently-written-entities ("Queries use the same syntax as search / filtering in the UI")
- [ ] `$allowEmpty` is not used anywhere: https://github.com/versionpress/versionpress/blob/c916d7d0fd25f97a52a18370a70966e803e3e104/plugins/versionpress/src/Utils/QueryLanguageUtils.php#L24

Tests:
- [ ] We could add some capitalized values into `entityMatchesRightQuery` and `entityDoesntMatchWrongQuery`.
- [ ] Revert case-insensitive `assertEquals()` from `queryLanguageUtilsCreatesCorrectRules` and `queryLanguageUtilsGeneratesCorrectGitLogQuery` and check if it works.
- [ ] Some tests from `QueryLanguageUtilsTest` and `EntityInfoTest` probably duplicates some assumptions now.

pavelevap:
Resolves #1439 

Case-insensitive comparison moved from `QueryLanguageUtils::createRulesFromQueries` and `QueryLanguageUtils::entityMatchesSomeRule` to `QueryLanguageUtils::createGitLogQueryFromRule`. This ensures that entities comparison will be case-sensitive and filtering commits case-insensitive without losing any information related to capitalized characters.

### DIFF
--- a/plugins/versionpress/src/Utils/QueryLanguageUtils.php
+++ b/plugins/versionpress/src/Utils/QueryLanguageUtils.php
@@ -226,16 +226,16 @@ class QueryLanguageUtils
             }
 
             if (substr($key, 0, 5) === 'x-vp-') {
-                $prefix = '';
+                $key = str_replace( 'x-vp-', 'X-VP-', $key );
             } else {
                 if (substr($key, 0, 3) === 'vp-') {
-                    $prefix = '\(X-\)\?';
+                    $key = '\(X-\)\?' . str_replace( 'vp-', 'VP-', $key );
                 } else {
-                    $prefix = '\(X-VP-\|VP-\)';
+                    $key = '\(X-VP-\|VP-\)' . $key;
                 }
             }
 
-            $query .= ' --grep="^' . $prefix . $key . ': \(' . implode('\|', $values) . '\)$"';
+            $query .= ' --grep="^' . $key . ': \(' . implode('\|', $values) . '\)$"';
         }
 
         return $query;
--- a/plugins/versionpress/src/Utils/QueryLanguageUtils.php
+++ b/plugins/versionpress/src/Utils/QueryLanguageUtils.php
@@ -225,10 +225,10 @@ class QueryLanguageUtils
             }
 
             if (substr($key, 0, 5) === 'x-vp-') {
-                $key = str_replace( 'x-vp-', 'X-VP-', $key );
+                $key = str_replace('x-vp-', 'X-VP-', $key);
             } else {
                 if (substr($key, 0, 3) === 'vp-') {
-                    $key = '\(X-\)\?' . str_replace( 'vp-', 'VP-', $key );
+                    $key = '\(X-\)\?' . str_replace('vp-', 'VP-', $key);
                 } else {
                     $key = '\(X-VP-\|VP-\)' . $key;
                 }
--- a/plugins/versionpress/src/Utils/QueryLanguageUtils.php
+++ b/plugins/versionpress/src/Utils/QueryLanguageUtils.php
@@ -21,10 +21,9 @@ class QueryLanguageUtils
      *  array = ['some_field' => ['value'], 'other_field' => ['other_value']]
      *
      * @param $queries
-     * @param $allowEmpty boolean Allow empty values
      * @return array
      */
-    public static function createRulesFromQueries($queries, $allowEmpty = false)
+    public static function createRulesFromQueries($queries)
     {
         $rules = [];
         foreach ($queries as $query) {
@@ -48,7 +47,7 @@ class QueryLanguageUtils
                 isset($match[4]) ? $match[4] : (
                 isset($match[3]) ? $match[3] : ''));
 
-                if ($value !== '' || $allowEmpty) {
+                if ($value !== '') {
                     if (!isset($ruleParts[$key])) {
                         $ruleParts[$key] = [];
                     }
--- a/plugins/versionpress/src/Utils/QueryLanguageUtils.php
+++ b/plugins/versionpress/src/Utils/QueryLanguageUtils.php
@@ -36,7 +36,7 @@ class QueryLanguageUtils
 
             $ruleParts = [];
             foreach ($matches as $match) {
-                $key = empty($match[2]) ? 'text' : strtolower($match[2]);
+                $key = empty($match[2]) ? 'text' : $match[2];
 
                 /* value can be in 3rd, 4th or 5th index
                  *
@@ -44,9 +44,9 @@ class QueryLanguageUtils
                  * 4th index => value is in double quotes
                  * 5th index => value is without quotes
                  */
-                $value = strtolower(isset($match[5]) ? $match[5] : (
+                $value = isset($match[5]) ? $match[5] : (
                 isset($match[4]) ? $match[4] : (
-                isset($match[3]) ? $match[3] : '')));
+                isset($match[3]) ? $match[3] : ''));
 
                 if ($value !== '' || $allowEmpty) {
                     if (!isset($ruleParts[$key])) {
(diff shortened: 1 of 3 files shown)


plugins/versionpress/src/Utils/QueryLanguageUtils.php
```diff
@@ -225,16 +225,16 @@ public static function createGitLogQueryFromRule($rawRule)
...
             if (substr($key, 0, 5) === 'x-vp-') {
-                $prefix = '';
+                $key = str_replace('x-vp-', 'X-VP-', $key);
             } else {
                 if (substr($key, 0, 3) === 'vp-') {
-                    $prefix = '\(X-\)\?';
+                    $key = '\(X-\)\?' . str_replace('vp-', 'VP-', $key);
                 } else {
-                    $prefix = '\(X-VP-\|VP-\)';
+                    $key = '\(X-VP-\|VP-\)' . $key;
```
pavelevap: I stayed with capitalized letters here and changed `queryLanguageUtilsGeneratesCorrectGitLogQuery` test accordingly to follow `VP-Action` example (also capitalized letters) and made some small changes to allowing test to pass (`assertEquals()` is now case sensitive). But it can be also switched to lowercase characters.

borekb:
> This ensures that entities comparison will be case-sensitive

Is that the goal though? I _think_ the intent of #1439 was to:

- Update query parsing so that it maintains the original letter casing, i.e., it doesn't lose any information when translating from YAML to PHP.
- Implement case insensitivity suitably for each scenario:
    - For Git searching, use `git log -i` flag.
    - For entity comparison, use `strtolowercase` in `entityMatchesSomeRule`.

Importantly, this would not change any user-facing behavior, just organize the code better and also fix the tests (for example, this change is highly desirable: https://github.com/versionpress/versionpress/pull/1444/files#diff-c2ff5a3b77d21156b05214cdb94eb655R107). If we keep case insensitivity as far as the user is concerned, many things in this PR will become simpler.

pavelevap:
> For entity comparison, use strtolowercase in entityMatchesSomeRule.

Yes, you are right, I see it now. I probably moved the goal a little bit for this part. During work on this PR I realized that we also do not need using `strtolower()` in `QueryLanguageUtils::entityMatchesSomeRule` and we can compare entities case-sensitive (and also avoid problems with missing multibyte string functions). It leads me to use `strtolower()` for array keys in `QueryLanguageUtils::createGitLogQueryFromRule` to fix converting rules for git log and remove restriction on case insensitivity for some `QueryLanguageUtilsTest`.

For Git searching is still used `git log -i` flag (checked manually), PR only fixes differences between `QueryLanguageUtils::createGitLogQueryFromRule` and related test. Also query parsing maintains the original letter casing now.

borekb:
Yes, this PR does many good things and I think it's about 80% where it should be. It just wasn't clear to me what the goal regarding entity matching was as the issue mentions docs and their necessary update around "queries use the same syntax as search / filtering in the UI" – https://github.com/versionpress/versionpress/issues/1439#issue-440051275.

BTW, I've updated my previous comment to highlight that `strtolower` is just an example; you're right that there might be better implementations.

plugins/versionpress/tests/Unit/EntityInfoTest.php
```diff
@@ -128,7 +128,7 @@ public function ignoredEntityCapitalizedIsCorrectlyIdentified()
         $entity = [
             'some_field' => 'b',
             'other_field' => 'a',
-            'capitalized_value_field' => 'VaLuE', // comparison must be case insensitive
```
borekb: This is how the test should look – we indeed want to test that if `VaLuE` is provided, it is still matched by a rule  that uses uppercase (or lowercase, or any-case) `VALUE`.

plugins/versionpress/tests/Unit/QueryLanguageUtilsTest.php
```diff
@@ -101,8 +104,7 @@ public function ruleAndQueryProvider()
     public function queryLanguageUtilsCreatesCorrectRules($query, $expectedRules)
     {
         $rules = QueryLanguageUtils::createRulesFromQueries($query);
-        // Perform case insensitive match
-        $this->assertEquals($expectedRules, $rules, '', 0, 10, false, true);
+        $this->assertEquals($expectedRules, $rules);
```
borekb: 👍  this is a very welcome change, basically making tests actually test something :)

plugins/versionpress/tests/Unit/QueryLanguageUtilsTest.php
```diff
@@ -178,12 +179,12 @@ public function rulesAndGitLogQueryProvider()
                 '-i --all-match --grep="^VP-Action: \(entity\|.*\)/.*/\(.*vp.*\|vpid\)$"'
             ],
             [['text' => ['text1', 'Test text', '*']], '-i --all-match --grep="text1" --grep="Test text" --grep=".*"'],
-            [['x-vp-another-key' => ['Test value']], '-i --all-match --grep="^x-vp-another-key: \(Test value\)$"'],
-            [['vp-another-key' => ['Test value']], '-i --all-match --grep="^\(x-\)\?vp-another-key: \(Test value\)$"'],
-            [['another-key' => ['Test value']], '-i --all-match --grep="^\(x-vp-\|vp-\)another-key: \(Test value\)$"'],
+            [['X-vp-another-key' => ['Test value']], '-i --all-match --grep="^X-VP-another-key: \(Test value\)$"'],
```
borekb: I think that no case juggling is necessary here. If the input is `X-vp-another-key`, Git search should be `... --grep="^X-vp-another-key` as well.

Similarly for the lines below.

pavelevap: Yes, but we need it for `QueryLanguageUtils::createGitLogQueryFromRule` and also this test (and I believe that it was the original reason to make tests case-insensitive before).
Example:  `vp-another-key` was automatically transformed into `X-vp-another-key` and test would fail because of expecting `x-vp-another-key`.
https://github.com/versionpress/versionpress/blob/c7695ab4cb573fc81376cb984b127657d3cec523/plugins/versionpress/src/Utils/QueryLanguageUtils.php#L231
Result will be now capitalized everytime: `X-VP-another-key` (see changes in `QueryLanguageUtils::createGitLogQueryFromRule`). But I am not sure if this is the best solution, see my previous comment: https://github.com/versionpress/versionpress/pull/1444#pullrequestreview-236416513.

plugins/versionpress/tests/Unit/EntityInfoTest.php
pavelevap: Yes, we are providing `VALUE` in original `$entitySchema`, so we are also expecting `VALUE`. Comparison is not case insensitive now, so `VaLuE` would fail.

pavelevap:
I am closing this PR, superseded by #1450.

PR #1448, opened 2019-05-14
Update CloneMergeTest and related to finish #1409

borekb:
Issue: #1409 
Follow-up to #1410 and #1421 

After we decided not to merge changes of #1421, see https://github.com/versionpress/versionpress/pull/1421#issuecomment-492123692, this is a follow-up PR to fix some of the comments and bring over some of the relevant commits from #1421, namely:

- Do not replace line endings on cloned site https://github.com/versionpress/versionpress/commit/099faf7d9787fccbd97f6ece10c7d635690b8a83 (originally https://github.com/versionpress/versionpress/pull/1421/commits/f07b6a610fc1faa145942563ed04daca7213e4fa).
- ~~TODO: should we bring over https://github.com/versionpress/versionpress/pull/1421/commits/9a84e45b7a57346e341ddbcd67c2e466a576f8ed as well?~~ Nope, see https://github.com/versionpress/versionpress/pull/1448#issuecomment-492163039.

### DIFF
--- a/plugins/versionpress/tests/Workflow/CloneMergeTest.php
+++ b/plugins/versionpress/tests/Workflow/CloneMergeTest.php
@@ -27,8 +27,7 @@ class CloneMergeTest extends PHPUnit_Framework_TestCase
 
         self::$cloneSiteConfig = self::getCloneSiteConfig(self::$siteConfig);
 
-        // Currently, workflow tests only pass when they are run on a fresh WordPress install, see #1409. That's why
-        // we force a full setup here. In the future, just `$wpAutomation->ensureTestSiteIsReady` should be used.
+        // Currently, we use an empty site as a basic test.
         $wpAutomation = new WpAutomation(self::$siteConfig, self::$testConfig->wpCliVersion);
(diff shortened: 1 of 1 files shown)


borekb:
@pavelevap, what do you think, should we update dates in End2End tests as per https://github.com/versionpress/versionpress/commit/9a84e45b7a57346e341ddbcd67c2e466a576f8ed or not?

I'm not sure: it probably does no harm but at the same time, now that End2End tests are separated from Workflow tests, maybe there's no reason to update E2E tests so that no entities are created within a single second.

pavelevap:
I am not sure, but there is no need to change predefined date to another predefined date now. Maybe I would remove all predefined dates as per https://github.com/versionpress/versionpress/pull/1421/commits/435df0af5084ea8c1a8e2cf583b843c6b50f7a61 (to better follow post sequence when examining container with tests), but it is also not needed now (and PR #1421 is not a complete list).

borekb:
Ok, thanks, so the PR is ready for review now.

PR #1450, opened 2019-05-15
Refactor case sensitivity in rule matching and commit searching, take 2

borekb:
Issue: #1439 

This PR performs several refactoring around how we treat case (in)sensitivity in `QueryLanguageUtils` and related tests. It doesn't change the public behavior – both Git history searching and entity matching are still case **in**sensitive.

In fact, this PR improves the behavior a bit so that not only values are case insensitive but also fields. For example, `FIELD: value` now matches `['field' => 'value']` (or any other combination of letter casing).

> Note: This PR replaces #1444 where most of the refactorings started, however, it also changed the behavior of entity matching to be case sensitive which we decided against for now.

Details:

- Query parsing in `createRulesFromQueries` now maintains original letter casing. For example, `field: VALUE` will be parsed to `['field' => 'VALUE']`.
- Entity matching in `entityMatchesSomeRule` now uses case insensitive comparison explicitly.
- Git history searching uses the `-i` flag for case insensitive behavior. 
    - Some smaller changes in `createGitLogQueryFromRule` were also necessary, see 244daf4f3d2fd88dd051943b96fe140f963d121b.
- Fields are also treated case insensitively now, see a1c91e87ce1454254bd8aec0cf2e11c5bfaea968.
- Tests around Git search no longer use an ugly workaround with `assertEquals($expected, $actual, '', 0, 10, false, true)` which actually hid some small problems in the test code. Tests have been updated to match the reality.

Still TODO:

- [x] https://github.com/versionpress/versionpress/pull/1450#issuecomment-492853541
- [x] https://github.com/versionpress/versionpress/pull/1450#issuecomment-492933179
- ~~Cherry-pick removal of `allowEmpty` from #1444 c9b4e4a82a6d96ed92720c1dee5294ffcff76f84~~ I'll do this separately, this PR should be about case sensitivity only.
- [x] Update public documentation on the case insensitive behavior, how it also depends on DB collation, etc. 517b76eb9fe8986618daaeaffed5a89c5597a3fd
- [x] Update regex101 example and code docs. https://regex101.com/r/wT6zG3/6 & https://github.com/versionpress/versionpress/pull/1450/commits/09fa3bc9d66bf9230c1494a46ec84e7e761d174f.

### DIFF
--- a/plugins/versionpress/src/Utils/QueryLanguageUtils.php
+++ b/plugins/versionpress/src/Utils/QueryLanguageUtils.php
@@ -76,7 +76,7 @@ class QueryLanguageUtils
             return ArrayUtils::all($rule, function ($values, $field) use ($entity) {
                 $field = strtolower($field);
 
-                if (!isset($entity[strtolower($field)])) {
+                if (!isset($entity[$field])) {
                     return false;
                 }
 
--- a/plugins/versionpress/src/Utils/QueryLanguageUtils.php
+++ b/plugins/versionpress/src/Utils/QueryLanguageUtils.php
@@ -70,10 +70,13 @@ class QueryLanguageUtils
      */
     public static function entityMatchesSomeRule($entity, $rules)
     {
+        $entity = array_change_key_case($entity, CASE_LOWER);
         return ArrayUtils::any($rules, function ($rule) use ($entity) {
             // check all parts of rule
             return ArrayUtils::all($rule, function ($values, $field) use ($entity) {
-                if (!isset($entity[$field])) {
+                $field = strtolower($field);
+
+                if (!isset($entity[strtolower($field)])) {
                     return false;
                 }
 
--- a/plugins/versionpress/src/Utils/QueryLanguageUtils.php
+++ b/plugins/versionpress/src/Utils/QueryLanguageUtils.php
@@ -2,6 +2,32 @@
 
 namespace VersionPress\Utils;
 
+/**
+ * This class groups several functions around a small query language that is used in search
+ * and for ignored & frequently written entities in plugin definitions.
+ *
+ * Rules of the language:
+ *
+ * 1. Basic search with wildcard support: `just text` (all words), `"just text" (exact match)`, `w*ldcards`.
+ * 2. `operator:value` syntax for operators, with optional spaces after the colon.
+ * 3. Logical OR for repeated operators: `operator:value1 operator:value2`.
+ * 4. Logical AND for multiple operators: `operator1:value operator2:value`.
+ * 5. Everything is case insensitivity: `"search term" author: Joe` is equivalent to `"SEarCH TErm" Author: joe`.
+ *
+ * Notes on case insensitivity:
+ *
+ * - Rules are parsed as they are, maintaining their original letter casing.
+ * - When constructing Git search command, rules are passed as is but an `-i` flag is added.
+ * - When matching rules for for ignored and frequently written entities, case insensitive comparisons are done.
+ * - When constructing SQL restrictions, for example, `(field LIKE "value")`, values are passed without any
+ *   case conversion and it's up to the configuration of MySQL how it treats that. By default, WordPress
+ *   sets the case insensitive `utf8_general_ci` collation.
+ *
+ * When updating this code, also think about our public documentation related to this, currently in:
+ *
+ * - searching-history.md
+ * - plugin-support.md
+ */
 class QueryLanguageUtils
 {
 
--- a/plugins/versionpress/src/Utils/QueryLanguageUtils.php
+++ b/plugins/versionpress/src/Utils/QueryLanguageUtils.php
@@ -14,6 +14,8 @@ namespace VersionPress\Utils;
  * 4. Logical AND for multiple operators: `operator1:value operator2:value`.
  * 5. Everything is case insensitivity: `"search term" author: Joe` is equivalent to `"SEarCH TErm" Author: joe`.
  *
+ * See examples in https://regex101.com/r/wT6zG3/6.
+ *
  * Notes on case insensitivity:
  *
  * - Rules are parsed as they are, maintaining their original letter casing.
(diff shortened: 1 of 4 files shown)


borekb:
Things I want to look at tomorrow:

- Do we want to keep the changes introduced in https://github.com/versionpress/versionpress/pull/1450/commits/a1c91e87ce1454254bd8aec0cf2e11c5bfaea968? It's a slight change in behavior, I'd argue for the better but it should be reviewed anyway.
- Do [these tests](https://github.com/versionpress/versionpress/blob/a1c91e87ce1454254bd8aec0cf2e11c5bfaea968/plugins/versionpress/tests/Unit/QueryLanguageUtilsTest.php#L69-L104) need enhancing? Maybe we're not covering all the real DB behaviors now, especially since the last commit.

borekb:
I'm also thinking about removing some recent changes to `EntityInfoTest`, for example, this one:

https://github.com/versionpress/versionpress/blob/a1c91e87ce1454254bd8aec0cf2e11c5bfaea968/plugins/versionpress/tests/Unit/EntityInfoTest.php#L22

It seems to me that proper place to test these things is `QueryLanguageUtilsTest` and we already have it here:

https://github.com/versionpress/versionpress/blob/a1c91e87ce1454254bd8aec0cf2e11c5bfaea968/plugins/versionpress/tests/Unit/QueryLanguageUtilsTest.php#L17

with this data provider:

https://github.com/versionpress/versionpress/blob/a1c91e87ce1454254bd8aec0cf2e11c5bfaea968/plugins/versionpress/tests/Unit/QueryLanguageUtilsTest.php#L35-L40

So the test code in `EntityInfoTest` feels duplicate.

plugins/versionpress/src/Utils/QueryLanguageUtils.php
```diff
@@ -70,10 +70,13 @@ public static function createRulesFromQueries($queries, $allowEmpty = false)
...
     public static function entityMatchesSomeRule($entity, $rules)
     {
+        $entity = array_change_key_case($entity, CASE_LOWER);
         return ArrayUtils::any($rules, function ($rule) use ($entity) {
             // check all parts of rule
             return ArrayUtils::all($rule, function ($values, $field) use ($entity) {
-                if (!isset($entity[$field])) {
+                $field = strtolower($field);
+
+                if (!isset($entity[strtolower($field)])) {
```
borekb: Oh, I forgot to remove `strtolower` here.

plugins/versionpress/tests/Unit/QueryLanguageUtilsTest.php
```diff
@@ -101,8 +101,6 @@ public function ruleAndQueryProvider()
     public function queryLanguageUtilsCreatesCorrectRules($query, $expectedRules)
     {
         $rules = QueryLanguageUtils::createRulesFromQueries($query);
-        // Perform case insensitive match
-        $this->assertEquals($expectedRules, $rules, '', 0, 10, false, true);
```
borekb: Ugh, I must have been sleeping already, this removes not only the case sensitivity hack but also the assert itself!

plugins/versionpress/src/Utils/QueryLanguageUtils.php
borekb: Resolved by 25150aadce41798555089ffaa6ae7e0357379288.

plugins/versionpress/tests/Unit/QueryLanguageUtilsTest.php
borekb: Resolved by 4b4353cd6ab90e913b5fb5b4db9b782b1bb1824a.

borekb:
I've updated the PR description, there are still some things to do but it's generally ready for review.

plugins/versionpress/tests/Unit/QueryLanguageUtilsTest.php
```diff
@@ -185,12 +185,14 @@ public function rulesAndGitLogQueryProvider()
                 '-i --all-match --grep="^VP-Action: \(entity\|.*\)/.*/\(.*vp.*\|vpid\)$"'
             ],
             [['text' => ['text1', 'Test text', '*']], '-i --all-match --grep="text1" --grep="Test text" --grep=".*"'],
+
             [['x-vp-another-key' => ['Test value']], '-i --all-match --grep="^x-vp-another-key: \(Test value\)$"'],
-            [['vp-another-key' => ['Test value']], '-i --all-match --grep="^\(x-\)\?vp-another-key: \(Test value\)$"'],
-            [['another-key' => ['Test value']], '-i --all-match --grep="^\(x-vp-\|vp-\)another-key: \(Test value\)$"'],
+            [['X-VP-another-key' => ['Test value']], '-i --all-match --grep="^X-VP-another-key: \(Test value\)$"'],
+            [['vp-another-key' => ['Test value']], '-i --all-match --grep="^\(X-\)\?vp-another-key: \(Test value\)$"'],
```
borekb: All these tests now verify the following behavior, implemented in 244daf4f3d2fd88dd051943b96fe140f963d121b:

- Prefix _detection_, i.e., whether a field already contains `X-VP-` or `VP-`, is case insensitive. 
- But, for actually producing the Git search string, there is no active conversion:
    - VersionPress always adds `(X-)VP-` (if necessary) in upper case.
    - User input is untouched.

This can lead to mixed casing like `X-VP-some-key` which is fine – the `-i` flag won't care.

borekb:
@pavelevap, this is now ready for review, I've pushed everything I wanted to.

plugins/versionpress/src/Utils/QueryLanguageUtils.php
```diff
@@ -2,13 +2,41 @@
...
+ * 4. Logical AND for multiple operators: `operator1:value operator2:value`.
+ * 5. Everything is case insensitivity: `"search term" author: Joe` is equivalent to `"SEarCH TErm" Author: joe`.
+ *
+ * See examples in https://regex101.com/r/wT6zG3/6.
+ *
+ * Notes on case insensitivity:
+ *
+ * - Rules are parsed as they are, maintaining their original letter casing.
+ * - When constructing Git search command, rules are passed as is but an `-i` flag is added.
+ * - When matching rules for for ignored and frequently written entities, case insensitive comparisons are done.
```
pavelevap: Small typo, double "for". I love this documentation ❤️

plugins/versionpress/tests/Unit/QueryLanguageUtilsTest.php
```diff
@@ -20,15 +20,24 @@ public function entityMatchesRightQuery($queries, $entity)
     public function validQueryAndEntityProvider()
     {
         return [
+        	// basic rule matching
```
pavelevap: Tabs/spaces

borekb:
Thanks for the review, @pavelevap.

PR #1455, opened 2019-06-15
change default icon

mustafauysal:
I didn't like the default icon at the top for the admin menu. `dashicons-backup` (or any other preferred icon https://developer.wordpress.org/resource/dashicons/) might be better?

### DIFF
--- a/plugins/versionpress/setup-hooks.php
+++ b/plugins/versionpress/setup-hooks.php
@@ -719,7 +719,7 @@ function vp_admin_menu()
         'manage_options',
         'versionpress',
         'versionpress_page',
-        null,
+        'dashicons-backup,
         0.001234987
     );
 
--- a/plugins/versionpress/setup-hooks.php
+++ b/plugins/versionpress/setup-hooks.php
@@ -719,7 +719,7 @@ function vp_admin_menu()
         'manage_options',
         'versionpress',
         'versionpress_page',
-        'dashicons-backup,
+        'dashicons-backup',
         0.001234987
     );
 


pavelevap:
Hi @mustafauysal,

thank you for this PR.

Proposed dashicon: https://developer.wordpress.org/resource/dashicons/#backup

Before: 
image

After:
image

I am not sure if we need to change default settings icon now, but maybe we could try new logo, see for example our Twitter account: https://twitter.com/versionpress (our branding is a little bit complicated, there is a hosting service and VersionPress plugin). @borekb

plugins/versionpress/setup-hooks.php
```diff
@@ -719,7 +719,7 @@ function vp_admin_menu()
         'manage_options',
         'versionpress',
         'versionpress_page',
-        null,
+        'dashicons-backup,
```
pavelevap: Missing apostrophe at the end.

borekb:
Thanks for posting the screenshots, @pavelevap. I actually like the change and would be inclined to accept it. (Sorting out the general icon / branding is a bigger task that we should definitely get to one day 😄.)

Thanks for the PR, @mustafauysal!

plugins/versionpress/setup-hooks.php
mustafauysal: Oh! my bad, this happens when jumping to file directly on GitHub 🤦🏻‍♂️ 
fixed in 1fb585263317c5c234749d9fb9dfd0b089a8dcfb

mustafauysal:
> Sorting out the general icon / branding is a bigger task that we should definitely get to one day 😄

You guys know, `add_menu_page()` supports base64-encoded SVG. Maybe converting from the exists twitter icon might not hard at all 🙃

PR #1458, opened 2019-07-09
feat: added prefix option

efuturetoday:
@borekb as discussed [in Gitter](https://gitter.im/versionpress/versionpress?at=5d249fafd0e06b6675746d03).

To make third party plugins that use a custom database prefix (different than `$wpdb->prefix`) compatible with VersionPress, you can now define a property `prefix` in `schema.yml` in order to prefix the table name of the entity with a custom one.

### DIFF
--- a/docs/content/en/developer/plugin-support.md
+++ b/docs/content/en/developer/plugin-support.md
@@ -207,7 +207,7 @@ Again, this is prefix-less; `wp_` or another prefix will be added automatically.
 If you want to define a custom prefix for the entity you can define the `prefix` property. When defined Versionpress will prefix the table with it instead of the default one configured in `wp-config.php` (mostly `wp_`).
 
 ```yaml
-b2sUsers:
+b2sUser:
   table: users
   prefix: b2s_
   ...
--- a/docs/content/en/developer/plugin-support.md
+++ b/docs/content/en/developer/plugin-support.md
@@ -204,6 +204,15 @@ post:
 
 Again, this is prefix-less; `wp_` or another prefix will be added automatically.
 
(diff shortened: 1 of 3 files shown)


borekb:
Hi @efuturetoday, thanks for the contribution!

Would you please also update `docs/content/en/developer/plugin-support.md` to describe this change to users? On an example, it will be easier to judge the naming of this (in the code, I see you used two terms for this, `prefix` and `tablePrefix`, so we should decide if we need two etc.).

efuturetoday:
Hey @borekb,
I updated the docs and introduced the feature.
Maybe we should name the property `prefix` in the class `DbSchemaInfo`to `defaultPrefix` to be more precise.

borekb:
Hi @efuturetoday, sorry for the delay, I was on holiday 🌴.

Thanks for describing the feature in the docs, you've done a good job with it 👍 .

## Do we want to support this feature?

That's the biggest question. The whole point of a table prefix (by that, I mean `$wpdb->prefix`) is that there can be multiple WordPress installations inside a single database; if a plugin doesn't respect that, it could be argued that it is badly coded.

Do we want to "pollute" our schema with things like that? It also leaks into code, see for example https://github.com/versionpress/versionpress/pull/1458/commits/46601e280d426f3a4463871bd8e378ca7f7e7ee6 which renames `prefix` to `defaultPrefix` – but there's no "default prefix" in WordPress so the code is slightly harder to understand for the general case.

I'm personally torn on this – on one hand, I can see how this feature would be useful for some websites (and we at VersionPress happily acknowledge that badly coded plugins are a reality that must be deal with) but also, I don't like making the `schema.yml` format more complex for the sake of workarounds.

Maybe a compromise would be to add a special section like this (pseudocode):

```yaml
b2sUser:
  table: users
  non-standard-features:
    hardcoded-table-prefix: b2s_
  ...
```

I'd sort of like to make bad things ugly.

What do you think? cc @pavelevap

Issue #71, opened 2015-10-26
Support WordPress auto-update


borekb:
_This is an imported issue, original reference WP-71, created on 15 Aug 2014._

When WordPress sees `.git` in the site root, it [turns off auto updates](https://codex.wordpress.org/Configuring_Automatic_Background_Updates). Possible solutions:
1. Leave default behavior and do not update WP
2. Force auto-update via wp-config.php
3. Force auto-update in a case where VersionPress created the .git folder

We'll probably do 3.

PR #1464, opened 2019-09-26
Always allow automatic updates

candrews:
`automatic_updates_is_vcs_checkout` should return false as long as VersionPress is active allowing automatic updates

Whether the first commit in the git repo is a VP activation commit or not is irrelevant.

@borekb @pavelevap What do you think?

### DIFF
--- a/plugins/versionpress/setup-hooks.php
+++ b/plugins/versionpress/setup-hooks.php
@@ -78,8 +78,7 @@ if (VersionPress::isActive()) {
 //----------------------------------
 
 add_filter('automatic_updates_is_vcs_checkout', function () {
-    $forceUpdate = UninstallationUtil::uninstallationShouldRemoveGitRepo(); // first commit was created by VersionPress
-    return !$forceUpdate; // 'false' forces the update
+    return !VersionPress::isActive(); // 'false' allows the update
 });
 
 function vp_register_hooks()


borekb:
Hi @candrews, thanks for the PR. I'm going to merge this as I think it improves things but am not 100% sure what the expected behavior should be. I think the idea for the original code was to only mess with auto-updates when we're certain that VersionPress created the Git repository (as indicated in #71) but now I agree that it's an arbitrary distinction.

Is it always desired to enable auto-updates when VersionPress is active? I'm not entirely sure but WordPress has other means like `WP_AUTO_UPDATE_CORE` or `auto_update_...` filters so there should be enough control for everybody to fit their needs. Allowing auto-updates by default is probably the sensible thing to do, so thank you, Craig, for implementing it.

plugins/versionpress/setup-hooks.php
```diff
@@ -78,8 +78,7 @@
 //----------------------------------
 
 add_filter('automatic_updates_is_vcs_checkout', function () {
-    $forceUpdate = UninstallationUtil::uninstallationShouldRemoveGitRepo(); // first commit was created by VersionPress
-    return !$forceUpdate; // 'false' forces the update
+    return !VersionPress::isActive(); // 'false' forces the update
```
borekb: I think we should use the word "allow" or "enable" instead of "force", what do you think?

```suggestion
    return !VersionPress::isActive(); // 'false' enables auto-updates
```

candrews: That's fine with me - I used the word "force" as that's what the original comment said :-)

borekb: Yep, we didn't choose the best word. If you agree, please commit the suggestion and with that, the PR will be ready to be merged.

candrews: Done!

borekb:
Thanks, @candrews!

PR #1471, opened 2019-12-22
Removed double slash in admin URL

emaildano:
Looks like there is two forward slashes. It's not breaking anything but it shouldn't be there.

Resolves #

Screen Shot 2019-12-22 at 10 07 18 AM


borekb:
Hi @emaildano, I don't think two slashes are normal, is that possibly something about your WP configuration?

